PR #1, opened 2020-05-08
Populating the repo with the old html guide files.

JesperIRL:
Adding the old guide files.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/guide pull/1/head:pull/1`
`$ git checkout pull/1`


src/producingChangeset.html
```diff
@@ -0,0 +1,485 @@
...
+
+<hr /></blockquote>
+
+<h2><a name="push">Pushing</a></h2>
+
+<p> In order to push changesets into the parent repository, some additional
+configuration is required.  The following sections describe the operations that
+will be performed by users with push access. </p>
+
+<h3><a name="sshGen">Generating an SSH Key</a></h3>
```
irisclark, line 233: I hope this or similar anchor stays around long term. I often use it when answering user questions.

src/changePlanning.html
```diff
@@ -0,0 +1,336 @@
...
+
+            <br/><br/>An entirely new test (or tests) may not be required. For
+            example, if the bug is an existing regression test failure, then
+            when fixing the bug you should just add the new bug ID to the list
+            of space-delimited bugs in the
+            <a href="http://openjdk.java.net/jtreg/tag-spec.html#INFORMATIONAL_TAGS">@bug tag</a>
+            of the failing regression test, even if the test did not need to
+            be updated.</td></tr>
+
+    <tr><td class="aligned"><a name="noreg">N</a></td>
```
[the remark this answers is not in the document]
JesperIRL: As long as we preserve these names in the markdown the links will continue to work.

src/producingChangeset.html
JesperIRL: The ATX headers that is recommended in other OpenJDK documents that use markdown automatically creates links to all headers. The default is to use the text from the header so the header above would be written as

>    ### Generating an SSH Key

which in turn would generate this html

<h3 id="generating-an-ssh-key">Generating an SSH Key</h3>

So all headers will have links but by default they won't be the same links as they used to be. I would prefer if we could stick with the defaults if possible.

irisclark: I agree.  We should use default links whenever possible.  My only concern is that the new links are stable.

PR #9, opened 2020-05-25
Remove version numbering

JesperIRL:
Removed release notes and version numbers from all pages. Added a footer which displays file modification date and time instead. The modification time is currently only placed at the bottom of each file, while the version number was present both at the top and at the bottom. This was made on purpose. Please leave a comment if you disagree with that decision. It's a simple makefile change to add it at the top as well.

---------
### Progress

### Reviewers
 * Magnus Ihse Bursie ([ihse](@magicus) - no project role)

### Download
`$ git fetch https://git.openjdk.java.net/guide pull/9/head:pull/9`
`$ git checkout pull/9`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,8 @@ build:
 	mkdir build
 
 build/%.html: src/%.md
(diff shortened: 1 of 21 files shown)


JesperIRL:
Since there is reluctance to add the date to the files I have updated the patch. The original discussion that lead to adding the date actually suggested adding the date or the change hash of each file. This new patch adds the change hash instead. It also adds a link to the change history for each file.

PR #10, opened 2020-05-26
Usign perl instead of sed

JesperIRL:
It turns out that different common versions of sed behaves differently so the suggestion here is to replace it with perl.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/guide pull/10/head:pull/10`
`$ git checkout pull/10`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,8 @@ build:
 	mkdir build
 
 build/%.html: src/%.md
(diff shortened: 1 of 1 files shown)


JesperIRL:
Added a change to use html 4 as pandoc output. There were issues validating the html 5 output.

PR #21, opened 2020-07-02
Filing bug, ProblemListing, Backing out

JesperIRL:
---------
### Progress

### Reviewers
 * [Igor Ignatyev](https://openjdk.java.net/census#iignatyev) (@iignatev - no project role) ⚠️ Review applies to 37cebc84a812b6cfc3a10dae740797f07db2fda5

### Download
`$ git fetch https://git.openjdk.java.net/guide pull/21/head:pull/21`
`$ git checkout pull/21`


src/next.md
```diff
@@ -7,6 +7,164 @@
...
+TBD
+
+## Filing a Bug
+
+When a new failure is found in the JDK a bug should be filed to describe and track the issue. Depending on your role in the OpenJDK you can either use the [Bug Report Tool](https://bugreport.java.com/) or, if you are Author in an OpenJDK Project, report the bug directly in the [JDK Bug System](https://bugs.openjdk.java.net/). Try to make the bug report as complete as possible to make it easier to triage and investigate the bug. It's not the reporter's responsibility to set a correct priority, but a good guess is always better than a default value. To help with setting the right priority consider things like how the bug impacts the product and our testing, how likely is it that the bug triggers, and how difficult is it to work around the bug if it's not fixed soon.
+
+A few things to keep in mind when filing a new bug:
+
+* Before filing a bug, verify that there isn't already a bug filed for this issue.
+* Add relevant labels like `intermittent`, `regression`, `testbug` etc.
```
iignatev, line 26: didn't we _deprecated_ `testbug` label and decided to use `noreg-self` to mark all test bugs?

src/next.md
```diff
@@ -7,6 +7,164 @@
...
+test/jdk/ProblemList-Xcomp.txt
+test/langtools/ProblemList.txt
+test/langtools/ProblemList-graal.txt
+test/lib-test/ProblemList.txt
+~~~
+
+Use the suitable ProblemList and add a line like this in the proper section:
+
+~~~
+sun.tools.jcmd.MyTest.java                        4711   windows-all
```
iignatev, line 69: this actually should be filename:
```suggestion
sun/tools/jcmd/MyTest.java                        4711   windows-all
```

src/index.md
```diff
@@ -25,6 +25,10 @@
   * [Merging](producingChangeset.html#merging)
   * [Pushing](producingChangeset.html#pushing)
 * [What Happens Next](next.html)
+  * [Requesting a Backport](next.html#requesting-a-backport)
+  * [Filing a Bug](next.html#filing-a-bug)
+  * [ProblemListing or Ignoring a test](next.html#problemlisting-or-ignore-ing-a-test)
```
iignatev, line 30: shouldn't it use the exact same words as the section's title?
```suggestion
  * [ProblemListing or `@ignore`-ing a Test](next.html#problemlisting-or-ignore-ing-a-test)
```

src/next.md
JesperIRL: Yes we did. You're absolutely right. Will change that.

src/next.md
JesperIRL: Right. Will fix.

src/index.md
JesperIRL: Good catch!

JesperIRL:
Thank you Igor! All comments fixed and I also added the example for ProblemListing single test cases that you sent.

src/next.md
```diff
@@ -66,27 +66,48 @@ test/lib-test/ProblemList.txt
...
 This would lead to `sun.tools.jcmd.MyTest.java` being ProblemListed only on `windows-all`. The proper way to write this is:
 
 ~~~
-sun.tools.jcmd.MyTest.java                        4710,4711   generic-all,windows-all
+foo/bar/MyTest.java                        4710,4711   generic-all,windows-all
 ~~~
 
 Although `windows-all` isn't strictly required in this example, it's preferable to specify platforms for each bugid (unless they are all `generic-all`), this makes it easier to remove one of the bugs from the list.
 
+#### ProblemListing some, but not all, test cases in a file
```
iignatev, line 90: actually, due to [CODETOOLS-7902712](https://bugs.openjdk.java.net/browse/CODETOOLS-7902712), you can't use `foo/bar/MyTest.java` entry to exclude all test cases, you have to list all of them explicitly.

src/next.md
```diff
@@ -7,6 +7,185 @@
...
+~~~
+
+Use the suitable ProblemList and add a line like this in the proper section:
+
+~~~
+foo/bar/MyTest.java                        4711   windows-all
+~~~
+
+In this example, `MyTest.java` is ProblemListed on windows, tracked by bug `JDK-4711`.
+
```
prrace, line 73: Not sure where in this review to insert this but there is a sensible policy that you should make sure the bug report is open. If you are submitting it that is easy to ensure but there have been cases where bugs are closed (without necessarily being fixed) and the issue remains PL'd. A legitimate case where this occurs if a bug is fixed in one line of development but the fix is not (yet:?) backported and I'm not sure how the "sensible policy" accounts for that.

src/next.md
```diff
@@ -7,6 +7,185 @@
...
+The fix for the main issue should remove the test from the ProblemList or remove the `@ignore` keyword from the test.
+
+#### Triage excluded issues
+
+After a failure is handled by excluding a test, the main JBS issue should be re-triaged and possibly given a new priority. This should be handled by the standard triage process. A test exclusion results in an outage in our testing. This outage should be taken into consideration when triaging, in addition to the impact of the bug itself.
+
+## Backing Out a Change
+
+If a change cause a regression that can't be fixed within reasonable time the best way to handle the regression can be to back out the change. Backing out means that the inverse (anti-delta) of the change is pushed to effectively undo the change in the repository. There are two parts to this task, how to do the bookkeeping in JBS, and how to do the actual backout in mercurial.
```
prrace, line 147: cause -> causes

src/next.md
```diff
@@ -7,6 +7,185 @@
...
+
+The backout is a regular change and will have to go through the standard code review process, but is considered a trivial change and thus it requires only one Reviewer and will avoid the 24h code review window, even for areas where stricter rules usually applies. The rationale is that a backout is usually urgent in nature and the change itself is automatically created by hg, and reviewed by the person who is performing the backout, so even though only one additional Reviewer is required the change will in practice get two reviews.
```
prrace, line 149: where is the 24h code review window discussed ?
It might be a sensible policy for non-urgent fixes but I've never seen it documented/codified and we've never had it as a rule in the client area.

src/next.md
```diff
@@ -7,6 +7,185 @@
...
+
+### How to work with JBS when a change is backed out
+
+#. Close the original JBS issue **(O)**.
+   * "Verify" the issue and choose "Fix Failed".
```
prrace, line 154: I have strong objections to fix failed ever being used and oppose it being recommended here. Unless the fixer and their reviewers completely failed at their job what you usually have is some other problem caused by the fix and the fix actually succeeded.

src/next.md
JesperIRL: Thanks for catching that! I've added a comment about this.

src/next.md
```diff
@@ -7,6 +7,185 @@
...
+* [Backing Out a Change](#backing-out-a-change)
+
+## Requesting a Backport
+
+TBD
+
+## Filing a Bug
+
+When a new failure is found in the JDK a bug should be filed to describe and track the issue. Depending on your role in the OpenJDK you can either use the [Bug Report Tool](https://bugreport.java.com/) or, if you are Author in an OpenJDK Project, report the bug directly in the [JDK Bug System](https://bugs.openjdk.java.net/). Try to make the bug report as complete as possible to make it easier to triage and investigate the bug. It's not the reporter's responsibility to set a correct priority, but a good guess is always better than a default value. To help with setting the right priority consider things like how the bug impacts the product and our testing, how likely is it that the bug triggers, and how difficult is it to work around the bug if it's not fixed soon.
+
```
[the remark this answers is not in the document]
JesperIRL: Added this to the text.

src/next.md
```diff
@@ -7,6 +7,185 @@
...
+  * command line information
+  * relevant information from the logs
+* Only set CPU and/or OS fields if the bug **ONLY** happens on that particular platform.
+* Always file separate bugs for different issues.
+  * If two crashes looks related but not similar enough to for sure be the same, it is easier to close a bug as a duplicate than it is to extract the relevant info from a bug to create a new one later.
+
+## ProblemListing or `@ignore`-ing a Test
+
+Sometimes tests break. It could be due to bugs in the test itself, or due to changed functionality in the code that the test is testing. While working on a fix, it can be useful to stop the test from being executed in everyone else's testing to reduce noise, especially if the test is expected to fail for more than a day. There are two ways to stop a test from being run in standard test runs: ProblemListing and using the `@ignore` keyword. Removing tests isn't the standard way to remove a failure. A failing test is expected to be a regression and should ideally be handled promptly with high urgency.
+
```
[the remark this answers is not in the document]
JesperIRL: Agreed. I have updated the text to reflect this.

src/next.md
```diff
@@ -7,6 +7,185 @@
...
+I'll say it right away so that it's not forgotten at the end: Remember to remove the JBS id from the ProblemList or the test when the bug is closed. This is especially easy to forget if a bug is closed as a duplicate or 'Will Not Fix'.
+
+### ProblemListing jtreg tests
+
+ProblemListing is used for a short term exclusion while a test is being fixed, and for the exclusion of intermittently failing tests that cause too much noise, but can still be useful to run on an ad-hoc basis. ProblemListing is done in the file `ProblemList.txt`. There are actually several ProblemList files to choose from. Their location and name hint about what area or feature each file belongs to. Each file has sections for different components. All ProblemList files complement each other to build the total set of tests to exclude in JTReg runs.
+
```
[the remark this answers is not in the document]
JesperIRL: "is" -> "should be" ?

src/next.md
JesperIRL: This is a very good point. I'm not sure how to express it though since this is meant to be a document for everyone, not only Oracle developers.

I'm not sure what you mean with regards to the backport. If a fix is made in mainline it's removed from the ProblemList in mainline. The backport should remove it from the ProblemList in the update release once it's pushed there. The one thing that could be problematic is the `problemlist` label since that is only supposed to be on the main bug. But on the other hand I don't think there is any automation that looks at the ProblemLists in the update trains so having the label tied to the mainline issue won't be a problem in practice I think.

src/next.md
JesperIRL: Fixed.

src/next.md
JesperIRL: That text hasn't been written yet. The 24h-rule is in use in the VM. I don't know if any other area use this rule. It doesn't apply to urgent fixes. Once we write that text it should specify if there are differences between areas. Whether it's needed or not depends, I guess, on how distributed the active area developers are over timezones.

src/next.md
iignatev: I think Phil is talking about a separate line of development of current feature release, i.e. `jdk/jdk` vs `jdk/client`, and not about `jdk/jdk` vs `11u`. let's say you have fixed `JDK-4711` in `jdk/client` repo, but your fix hasn't been propagated to `jdk/jdk` yet, and folks in `jdk/jdk` are still affected by `JDK-4711` and see test failures (from non-problem-listed tests), so they want to go and problem-list them, but JBS says that `JDK-4711` is already closed.

src/next.md
iignatev: given 24h rule is VM-specific, I think it might be reasonable to omit in this text and just say that backouts are considered trivial changes, and have the difference in rules for trivial and non-trivial changes described in yet-not-written reviewing doc (or its VM part)

src/next.md
JesperIRL: I guess this is a question for those who normally handle fix verification and may have scripts that look for different verifications.

There's only four values to choose from: "None", "Verified", "Not verified", and "Fix failed".
"Verified" means that the fix solved the problem and no more action is required, so this is clearly not right. "Not verified" seems wrong since it actually was verified that the fix caused problems - or it wouldn't need to be backed out. "None" could be used in my mind, but I can imagine that there are filters that treats "None" as issues that needs verification. So changing to using this would probably cause problems. That leaves "Fix failed".

Maybe Joe knows why this was designed as it is?  Anyhow, it is the current process and we need to bring it up with the right people before changing it.

iignatev: AFAIK, The verification process isn't part of any OpenJDK process and is used/done mainly internally by Oracle. why do you think that verification status should be set for all backed out issues?

src/next.md
JesperIRL: Ah, of course. There is no implemented solution for that right now. Two fairly easy options would be 1, Merge client into mainline, or 2, use a different fix version for client like done for project repos. This is something that we should bring up in a different thread with a different audience.

src/next.md
JesperIRL: Yes, I agree. There is no need to define (again) what trivial means here. The word should be a link to where "trivial" is defined.

src/next.md
JesperIRL: The text is based on an ancient definition of how to back out changes. That text was JDK wide (not only HotSpot) but yes, internal to Oracle. If verification is not important or done for all issues I'd be happy to remove that sentence.

JesperIRL: There has been a longer discussion in the email thread on guide-dev that for some reason hasn't been copied here. From that and other discussions I now have a directive on this issue; The verification field is both useful and used, also for non-Oracle issues. It's not an Oracle internal field and anyone in the community can use it - it's actually documented on the OpenJDK wiki even though it's not always easy to find information in a wiki. For this reason it's actually even more important to clarify how it's intended to be used in the cases where we use it. Therefore we should include the verification part in this description as it is part of the process.

As for using "Fix failed"; this was very well explained by Igor in the same mail thread: this is the value that means that the bug still exists - and it does since the fix was backed out. It doesn't mean that the fix itself was bad or broken.

iignatev: just for the record, the email thread @JesperIRL referred to starts [here](https://mail.openjdk.java.net/pipermail/guide-dev/2020-July/000192.html).

AlanBateman:
Filing a Bug - are you going to say anything about checking if the issue exists in the main line, esp. with reproducible issues? Also what about advice on attaching a reproducer if possible?

"ProblemListing or `@ignore`-ing a Test".  The ProblemList.txt files are jtreg exclude list. Can this section be renamed to "How to exclude tests"?  I don't think we've used "@ignore" for long time and might it simpler to just not mention it.

iignatev:
Hi @AlanBateman,
> I don't think we've used "@ignore" for long time and might it simpler to just not mention it.

`@ignore` is still used, albeit not very often, in hotspot tests for cases similar to the ones described here; given we are seldomly running ProblemList-ed tests, it would be really unfortunate to have include tests which remove `/etc/shadow` in such runs.

JesperIRL:
> Filing a Bug - are you going to say anything about checking if the issue exists in the main line, esp. with reproducible issues? Also what about advice on attaching a reproducer if possible?

Good suggestion! I have added both.

> "ProblemListing or `@ignore`-ing a Test". The ProblemList.txt files are jtreg exclude list. Can this section be renamed to "How to exclude tests"? I don't think we've used "@ignore" for long time and might it simpler to just not mention it.

I agree with changing the header of this section, and I do like "How to"-sections, but currently the guide doesn't have this style so I suggest simply "Excluding a test".

As for the usage of `@ignore`; the keyword is used in ~100 tests, about a third in each of hotspot, jdk, and langtools, so it's clearly used. I've added a few words to say that this is a less common use case, but I think the description as such should remain.

iignatev:
> As for the usage of `@ignore`; the keyword is used in ~100 tests, about a third in each of hotspot, jdk, and langtools, so it's clearly used. I've added a few words to say that this is a less common use case, but I think the description as such should remain.

for the record, the discussion has continued on [the mailing list](https://mail.openjdk.java.net/pipermail/guide-dev/2020-July/000225.html) (but due to some reasons hasn't been bridged to this PR), [8249618](https://bugs.openjdk.java.net/browse/JDK-8249618) has been filed to normalize/clean up usage of `@ignore` vs ProblemList.

src/next.md
```diff
@@ -7,6 +7,189 @@
...
+A few things to keep in mind when filing a new bug:
+
+* Before filing a bug, verify that there isn't already a bug filed for this issue.
+* Add relevant labels like `intermittent`, `regression`, `noreg-self` etc.
+* Set affects version to the JDK version(s) where the failure was seen.
+  * If the failure is found in an update train of the JDK (e.g. 11.0.x), please make an effort to see if the bug is also present in [mainline](https://hg.openjdk.java.net/jdk/jdk/).
```
kevinrushforth, line 28: This isn't really a comment on your proposed change. I presume there will be a follow-on task to update the docs after next week when the jdk repo transitions from Mercurial to Git?

AlanBateman:
One other comment on this is that "Filing a Bug" asks for labels to be added. Is there anywhere to link to that describes these labels? If not, then maybe this section could at least include a sentence about the "regression" label as it's a really useful indication that the reporter thinks there is a regression. If possible, the reporter should try to narrow down which release or build the regression first appeared as this can save a lot of time when looking at a bug report.

JesperIRL:
> One other comment on this is that "Filing a Bug" asks for labels to be added. Is there anywhere to link to that describes these labels?

There is the JBS label dictionary that I intend to move into the guide. It's probably wise to do that sooner rather than later. Then each label can link to the description of that label. (https://wiki.openjdk.java.net/display/HotSpot/JBS+Label+Dictionary)

> If possible, the reporter should try to narrow down which release or build the regression first appeared as this can save a lot of time when looking at a bug report.

Yes, that's a very good point. I've added a sentence about that.

src/next.md
```diff
@@ -7,6 +7,190 @@
...
+* Always file separate bugs for different issues.
+  * If two crashes looks related but not similar enough to for sure be the same, it is easier to close a bug as a duplicate than it is to extract the relevant info from a bug to create a new one later.
+
+## Excluding a Test
+
+Sometimes tests break. It could be e.g. due to bugs in the test itself, due to changed functionality in the code that the test is testing, or changes in the environment where the test is executed. While working on a fix, it can be useful to stop the test from being executed in everyone else's testing to reduce noise, especially if the test is expected to fail for more than a day. There are two ways to stop a test from being run in standard test runs: ProblemListing and using the `@ignore` keyword. Removing tests isn't the standard way to remove a failure. A failing test is often a regression and should ideally be handled with high urgency.
+
+I'll say it right away so that it's not forgotten at the end: Remember to remove the JBS id from the ProblemList or the test when the bug is closed. This is especially easy to forget if a bug is closed as a duplicate or 'Will Not Fix'.
+
```
prrace, line 47: I think @iignatev pushed a jcheck update that will fail if the bug in the commit matches one in the problem list and it isn't being removed in the commit. So maybe we can add a note about that.

JesperIRL: Yes, good point! I have added a sentence about that.

src/next.md
JesperIRL: Yes, there is a Skara bug filed to update the guide, https://bugs.openjdk.java.net/browse/SKARA-652

irisclark (approved):
Thanks for adding this information to the Guide and patiently working will reviewers to address their concerns.

PR #26, opened 2020-09-01
Added JBS Label Dictionary

JesperIRL:
Moving the JBS label Dictionary from the OpenJDK wiki into the Guide. Did some updates of the text.

I was unable to get the layout I wanted (or any readable layout really) using plain markdown, so this page uses an html table. If anyone can achieve the same or very similar layout using plain markdown I'd be happy to remove the html.

I'm guessing this dictionary is missing several labels used by libraries and client since it has a hotspot background. Please let me know what labels to add (or just add them).

I also moved the noreg-* description from "Change Planning" into this dictionary as I think it fits better here. All labels have anchors so you can link directly to any label in the dictionary.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/guide pull/26/head:pull/26`
`$ git checkout pull/26`

### DIFF
--- a/src/changePlanning.md
+++ b/src/changePlanning.md
@@ -57,11 +57,11 @@ also imply "implementation for an enhancement".
 
 #. **Does a bug id exist for the work?**
 
-   ---  -----------------------------------------------------------------------
-    Y   Continue
+   **Yes**
+   :   Continue
 
-    N   Create a bug in an appropriate component/sub-component.
-   ---  -----------------------------------------------------------------------
+   **No**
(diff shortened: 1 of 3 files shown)


edvbld:
Please ignore this comment, it is only for debugging Skara's e-mail bridging.

irisclark (commented):
This consolidation if label knowledge/use is great!  I like the fact that there are anchors for each label and, in some cases, groups of related labels.

irisclark (approved):

PR #46, opened 2021-03-15
Improve and modernize the makefile

magicus:
A bunch of fixes to the make file, to use more modern/robust syntax, and to keep intermediate build results.

This is a prerequisite for any further changes to the make file.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed

### Reviewers
 * [Jesper Wilhelmsson](https://openjdk.java.net/census#jwilhelm) (@JesperIRL - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/guide pull/46/head:pull/46`
`$ git checkout pull/46`

PR #47, opened 2021-03-15
Split up monolithic markdown

magicus:
The original Developer's Guide was split into one html page per chapter. This was (rightly so) combined into a single html page previously by Jesper.

However, this also means that we got a single, monolithic markdown file to edit. This makes the file unnecessarily hard to navigate and edit.

This patch will instead split up the original index.md file into separate markdown documents, one per chapter. At build time, they are recombined into a single index.md file again (exactly identical with the one I started with).

This patch depends on https://github.com/openjdk/guide/pull/46.

---------
### Progress
- [ ] Change must be properly reviewed

### Download
`$ git fetch https://git.openjdk.java.net/guide pull/47/head:pull/47`
`$ git checkout pull/47`


JesperIRL:
Continuing the erroneously placed discussion from PR #46:
> I don't mind this change but...

If we want to future-proof this and make it a stable foundation we shouldn't have chapter numbers in the file names. I don't want to rename all chapters just because a new one is inserted or removed somewhere. Also, since we are quite early in the making of this document it's likely that chapters will move within the guide as well.

magicus:
I thought it was a simple and elegant solution to get the chapters to line up, both when combining them in the make file, and in the IDE when editing them (as long as your IDE sorts files based on name). Again, I don't see renaming as a big problem, maybe that's what makes our opinions differ on this patch.

But sure, if you really prefer, I can provide another way of telling the build system how to organize the files when combining them.

Does it sound like a good idea to you to list them in order in a text file?

magicus:
The world was not ready for this change just yet. :-) Withdrawing PR.

PR #62, opened 2021-07-08
Section about the release process

JesperIRL:
Answering some common questions around the release process.

### DIFF
--- a/src/index.md
+++ b/src/index.md
@@ -1741,6 +1741,17 @@ Throughout the release there are a number of milestones and phases that defines 
 
 Please note that the priority of a bug doesn't change just because you want to get your fix in late in the release, or if you want to be able to defer it. The priority is based on the severity of the bug and if it was deemed to be a P2 before, you better have a really good explanation to why that conveniently has changed by the end of the release. Being hard to fix is **not** a reason to lower the priority of a bug.
 
+### Milestone cheat sheet
+
+Even though there's nothing explicitly written in the process about deferring P1 and P2 bugs during the normal development phase, the assumption is that these aren't deferred unless time runs out at the end of the release cycle.
+
+|           | P1           | P2           | P3           | P4/P5        | Doc/Test     | Enhancement  |
+|:----------|:-------------|:-------------|:-------------|:-------------|:-------------|:-------------|
+| Dev phase | Fix          | Fix          | Fix or Defer | Fix or Defer | Fix or Defer | Fix or Defer |
+| RDP1      | Fix or Defer with approval | Fix or Defer with approval | Fix or Defer | Defer | Fix or Defer | Fix with approval or Defer |
+| RDP2      | Fix with approval or Defer with approval | Fix with approval or Defer with approval | Defer | Defer | Fix or Defer | Defer |
+| RC        | Fix with approval or Defer with approval | Defer | Defer | Defer | Defer    | Defer    |
+
 ## Forward ports
 
 During the rampdown of a release there are two repositories in play, the stabilization fork for the outgoing release, and the mainline repository where the next release is being developed. Any bugfix going into the stabilization fork is likely to be desired in mainline as well. As a developer you should **only** push your fix to the stabilization fork if you intend for it to go to both repositories. Your fix will be forward ported to mainline.
--- a/src/index.md
+++ b/src/index.md
@@ -1693,6 +1693,60 @@ reverse effect of earlier changeset
     merged result is left uncommitted.
 ~~~
 
+# JDK Release process
+
+::: {.box}
+[Quick Links]{.boxheader}
+
+* [JEP 3: JDK Release Process](https://openjdk.java.net/jeps/3)
+:::
+
+The JDK project has a well defined release process. [JEP 3](https://openjdk.java.net/jeps/3) describes this process in detail. This section tries to clarify some topics that often cause questions.
+
+## Release cycle
+
+The release cycle starts when development of a new release begins and ends when that release is delivered to the public. The current release cadence is six months. This means that every six months we start development of a new release, and every six months a new release is delivered. However, this doesn't mean that each release cycle is six months. As described below, the total development time for a release (the release cycle) is actually nine months. Obviously this in turn doesn't mean that all features are developed in nine months. Most features are developed for much longer time than that, and goes through long time development in other project repositories, and through a series of preview and experimental stages. But any feature that is to be included in a specific release has a specific window of nine months to integrate the code into mainline and fix all the remaining bugs.
+
+It may be tempting to integrate a new feature near the end of a release cycle, to get more time to fix all those last bugs before integration. Please don't. If you are getting close to the end of a release and you still just have one more bug to fix, please defer your feature to the next release. It's only six months out. Not only will this vouch for your new feature to be more stable on release, you will also help keeping the JDK as a whole more stable by allowing others to find and fix bugs in their new code that might come as a result of your changes.
+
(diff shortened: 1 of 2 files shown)


src/index.md
```diff
@@ -1693,6 +1693,71 @@ reverse effect of earlier changeset
...
+
+Integrating early in a release is preferable, but all new features can't be integrated at the same time. If many large changes enters the repository at the same time it will be more difficult to determine which change that caused all the new bugs. If you're about to integrate a larger change you must therefore communicate this on the relevant [mailing lists](#mailing-lists) to synchronize with other projects that may also be planning to integrate something soon.
+
+## Milestones and phases
+
+Throughout the release there are a number of milestones and phases that define where in the release cycle we are.
+
+[**The start of a release**]{#release-start}
+:    Since development is always ongoing in the mainline repository ([openjdk/jdk](https://github.com/openjdk/jdk)), the start of a new release can be said to begin when the former release is forked of from mainline. After the start of the release follows six months of development to implement and integrate all the cool stuff that will go into the next release. After these six months rampdown begins.
```
stuart-marks, line 1719: wording: probably "forked from the mainline" at the end of the first sentence.

src/index.md
```diff
@@ -1693,6 +1693,71 @@ reverse effect of earlier changeset
...
+
+[**Ramp Down Start (RDS)**]{#rds}
+:    The ramp down of a release starts with a fork of the mainline repository. A clone of the entire code base is made and hosted in a separate ramp down repository (e.g. [openjdk/jdk17](https://github.com/openjdk/jdk17)). During the ramp down of a release we focus on bug fixing and stabilization in order to get the JDK ready for release.
+
+[**Ramp Down Phase 1 (RDP1)**]{#rdp1}
+:    At RDS we enter RDP1. During this phase you may continue to fix P1-P3 product bugs. P4 and P5 product bugs should be deferred at this point. Test bugs (labeled `noreg-self`) and documentation bugs (labeled `noreg-doc`) can still be fixed in RDP1 regardless of their priority. To fix an enhancement an approval is required. See the [Late-Enhancement Request Process](https://openjdk.java.net/jeps/3#Late-Enhancement-Request-Process) for details on how to do that. If you want to defer a P1 or P2 bug during RDP1 you will also need an approval. See the [Bug-Deferral Process](https://openjdk.java.net/jeps/3#Bug-Deferral-Process) for more details.
```
stuart-marks, line 1725: It would be good to avoid duplication with JEP 3. Some of the text here seems to do that, such as late enhancements and bug deferrals. By contrast, the discussion in the previous section about when to put features into a release is indeed suitable for the Guide.

I'm not familiar with "RDS". It's not mentioned anywhere in JEP 3, nor have I heard it much in our discussions. Can't we just say that the mainline repository is forked at the beginning of RDP1, and the new rules apply to the stabilization repo?

src/index.md
```diff
@@ -1693,6 +1693,71 @@ reverse effect of earlier changeset
...
+
+[**Feature Complete (FC)**]{#fc}
+:    Feature Complete is declared when all features that are supposed to be in the release have been integrated into the release. With the six-month release cadence, FC is defined as a date rather than a set of features, so the features that are integrated by FC are the features that should be in the release. FC is normally the same day as RDS.
```
stuart-marks, line 1728: I'm not sure talking about "Feature Complete" is helpful. In projects that are feature-driven, "feature complete" means "we've completed all the features that have been committed for this release". That clearly doesn't apply here. So then you have to explain that it means something different. It doesn't seem to be a separate phase or milestone. The feature set is frozen at the beginning of RDP1 (with some exceptions permitted, namely, the late-enhancement requests). Put another way, the deadline for integrating JEPs and enhancements is the beginning of RDP1.

src/index.md
```diff
@@ -1693,6 +1693,71 @@ reverse effect of earlier changeset
...
+### Milestone cheat sheet
+
+Even though there's nothing explicitly written in the process about deferring P1 and P2 bugs during the normal development phase, the assumption is that these aren't deferred unless time runs out at the end of the release cycle.
+
+|           | P1           | P2           | P3           | P4/P5        | Doc/Test     | Enhancement  |
+|:----------|:-------------|:-------------|:-------------|:-------------|:-------------|:-------------|
+| Dev phase | Fix          | Fix          | Fix or Defer | Fix or Defer | Fix or Defer | Fix or Defer |
+| RDP1      | Fix or Defer with approval | Fix or Defer with approval | Fix or Defer | Defer | Fix or Defer | Fix with approval or Defer |
+| RDP2      | Fix with approval or Defer with approval | Fix with approval or Defer with approval | Defer | Defer | Fix or Defer | Defer |
+| RC        | Fix with approval or Defer with approval | Defer | Defer | Defer | Defer    | Defer    |
```
stuart-marks, line 1753: Is this table the same as in JEP 3?

src/index.md
```diff
@@ -1693,6 +1693,71 @@ reverse effect of earlier changeset
...
+
+## Forward ports
+
+During the rampdown of a release there are two repositories in play, the stabilization fork for the outgoing release, and the mainline repository where the next release is being developed. Any bugfix going into the stabilization fork is likely to be desired in mainline as well. As a developer you should push your fix to the stabilization fork **only**, even if you intend for it to go to both repositories. Your fix will be forward ported to mainline.
+
+All fixes that are pushed to the stabilization fork are forward ported to mainline. If you have a fix that is only intended for the stabilization fork you will have to manually back it out from mainline once it has been forward ported.
```
stuart-marks, line 1759: Is this the way it really works? Certainly every commit is pulled from the stabilization fork into the mainline, since we want the mainline to include all the commits from the previous release. But are all those commits necessarily merged onto the master branch? Offhand I can't think of a better way, but maybe this deserves more discussion. At the very least, maybe there needs to be some workflow. For example, if you want to fix JDK-xxx in the stabilization fork but not in the mainline, you need to file JDK-yyy to back out the fix after it's merged into the mainline, and then link the bugs together in a particular way, or something.

stuart-marks (changes requested):

src/index.md
JesperIRL: Fixed

src/index.md
JesperIRL: I removed RDS and reworked RDP1. Now referring to JEP 3 instead of repeating all the details.

src/index.md
JesperIRL: I removed FC.

src/index.md
JesperIRL: This is the way it works. I forward port all changes from JDK 17 to mainline, and if a change is not for mainline it must be backed out. I added a paragraph with your example to clarify the JBS part.

stuart-marks: OK. If this is the way it really works, then clearly that reality needs to be documented. I'm a bit surprised about this; on the other hand, I thought of several alternatives, but they're either unworkable or more complicated and don't offer any advantages. This is probably a pretty rare case, so maybe the workflow is fine as it stands.

The filing-two-bugs material looks reasonable. I might strengthen it to say something like the following.

« As soon as you know that there is a fix that needs to go into the stabilization fork but not the mainline, you should do the following: file a bug JDK-yyy to cover the backout work, link it to JDK-xxx, set its Fix Version to the next release, add a comment describing the situation, and set the priority to be relatively high (e.g., P3). Then, you have to wait until the JDK-xxx fix is forward ported to the mainline before actually fixing JDK-yyy. Making these settings in JDK-yyy will help ensure that it won't be missed. »

stuart-marks:
Updated changes all look good, thanks.

JesperIRL:
Thank you for your reviews @stuart-marks!

src/index.md
JesperIRL: I added your text, but as a list.

kevinrushforth (approved):

irisclark (approved):

PR #66, opened 2021-11-20
Section on backporting

JesperIRL:
Guidance on backporting.
Also moved "Backing out a change" to the testing section where I believe it fits better. This section is unchanged.

---------
### Progress

### Reviewers
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - Author) ⚠️ Review applies to afbb243214b48fed3f6c27bfac0dcca994d2036c
 * [Kevin Rushforth](https://openjdk.java.net/census#kcr) (@kevinrushforth - no project role) ⚠️ Review applies to d5f7a7fcd1c3ff922bb46733f48bdf48ea843dc4
 * [Magnus Ihse Bursie](https://openjdk.java.net/census#ihse) (@magicus - Author) ⚠️ Review applies to d5f7a7fcd1c3ff922bb46733f48bdf48ea843dc4
 * [Christoph Langer](https://openjdk.java.net/census#clanger) (@RealCLanger - no project role) ⚠️ Review applies to ab420f92afa37f70e36e08d6f23a17023416e1a7
 * [Goetz Lindenmaier](https://openjdk.java.net/census#goetz) (@GoeLin - no project role) ⚠️ Review applies to 9779db8ad5ff35078805b7bd6106b609d3ec3760

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/guide pull/66/head:pull/66` \
`$ git checkout pull/66`

`$ git checkout pull/66` \
`$ git pull https://git.openjdk.java.net/guide pull/66/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 66`

`$ git pr show -t 66`

</details>

<a href="https://git.openjdk.java.net/guide/pull/66.diff">https://git.openjdk.java.net/guide/pull/66.diff</a>

</details>


src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+-# Welcome to my modified JDK!
++# Welcome to the JDK!
+
+ For build instructions please see the
+ [online documentation](https://openjdk.java.net/groups/build/doc/building.html),
+~~~
+
+### How to work with mercurial when a change is backed out
+
+In order to backout a change, the `hg backout` command is recommended, which essentially applies the anti-delta of the change. Make sure you perform the backout in the most upstream repostiory the change has escaped to.
```
dholmes-ora, line 1450: typo: repostiory

dholmes-ora (approved):
Hi Jesper,

Changes seem reasonable to me.

Thanks,
David

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+
+::: {.box}
+[Terminology]{.boxheader}
+
+Main issue - The top issue in a backport hierarchy. Eg. [JDK-8272373](https://bugs.openjdk.java.net/browse/JDK-8272373) is a main issue, while [JDK-8277498](https://bugs.openjdk.java.net/browse/JDK-8277498) and [JDK-8277499](https://bugs.openjdk.java.net/browse/JDK-8277499) are backport issues of this main issue.
+:::
+
+In general there is no need to create backport issues in JBS manually. All work that is done in JBS in preparation for a backport (requesting approvals etc) is done in the main issue. The backport issue will be created automatically by the bots when you integrate the change to the source code repository.
+
+There can be cases where it is motivated to create a backport issue before the fix is done. In these cases set the fix version of the backport to `N-pool`, where `N` is the release train the backport is targeting. E.g. `17-pool`
```
kevinrushforth, line 1492: You might want to add a sentence saying that if a backport record is created ahead of time, all work done in JBS is still done in the main issue.

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
 
 After a failure is handled by excluding a test, the main JBS issue should be re-triaged and possibly given a new priority. This should be handled by the standard triage process. A test exclusion results in an outage in our testing. This outage should be taken into consideration when triaging, in addition to the impact of the bug itself.
 
+## Backing Out a Change
+
+If a change causes a regression that can't be fixed within reasonable time the best way to handle the regression can be to back out the change. Backing out means that the inverse (anti-delta) of the change is pushed to effectively undo the change in the repository. There are two parts to this task, how to do the bookkeeping in JBS, and how to do the actual backout in git or mercurial.
+
+The backout is a regular change and will have to go through the standard code review process, but is considered a [trivial](#trivial) change. The rationale is that a backout is usually urgent in nature and the change itself is automatically generated. In areas where two reviewers are normally required, only one additional Reviewer is required for a backout since the person who is performing the backout also will review the change.
```
irisclark, line 1377: Recommend that both instances of "reviewer*" in the final sentence be of the same case.  I would choose upper.

src/index.md
JesperIRL: Thank you for reviewing! Fixed the typo.

src/index.md
JesperIRL: Thank you for reviewing! Added the sentence.

src/index.md
JesperIRL: The first "reviewers" does not refer to two Reviewers. For instance in Hotspot two reviewers are required but only one of them have to be a Reviewer.

irisclark: Ah! Thanks for the clarification.  The existing case make sense.

kevinrushforth (approved):

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+
+Setting the fix version of a backport to `N` is always wrong. JDK `N` has already been released (or you wouldn't be backporting to it) and can't get any more fixes.
+
+## Requesting approvals for backports
+
+In order to be allowed to push a change to one of the OpenJDK Update Release repositories, an approval is required. The [official process for how to request push approval for a backport](https://openjdk.java.net/projects/jdk-updates/approval.html) describes in detail how to work with JBS when requesting approvals. In short, there's a label `jdk<release>u-fix-request` that should be added to the main JBS issue. Also put a motivation as to why the issue needs to be backported as a comment in the main issue. Once the label and motivation has been added, wait for the maintainers of the release to approve your request.
```
stefank, line 1500: "wait for the maintainers of the release to approve your request" - Should this section describe that they do so by adding the label `jdk<release>u-fix-yes`?

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+
+::: {.box}
+[Terminology]{.boxheader}
+
+Main issue - The top issue in a backport hierarchy. Eg. [JDK-8272373](https://bugs.openjdk.java.net/browse/JDK-8272373) is a main issue, while [JDK-8277498](https://bugs.openjdk.java.net/browse/JDK-8277498) and [JDK-8277499](https://bugs.openjdk.java.net/browse/JDK-8277499) are backport issues of this main issue.
+:::
+
+In general there is no need to create backport issues in JBS manually. All work that is done in JBS in preparation for a backport (requesting approvals etc) is done in the main issue. The backport issue will be created automatically by the bots when you integrate the change to the source code repository.
+
+There can be cases where it is motivated to create a backport issue before the fix is done. In these cases set the fix version of the backport to `N-pool`, where `N` is the release train the backport is targeting. E.g. `17-pool`. Please note that even if a backport issue is created ahead of time, all work done in JBS is stil done in the main issue.
```
magicus, line 1492:
```suggestion
There can be cases where it is motivated to create a backport issue before the fix is done. In these cases set the fix version of the backport to `N-pool`, where `N` is the release train the backport is targeting. E.g. `17-pool`. Please note that even if a backport issue is created ahead of time, all work done in JBS is still done in the main issue.
```

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+
+## Using the Skara tooling to help with backports
+
+The Skara project developed support for backports. [The official Skara documentation](https://wiki.openjdk.java.net/display/SKARA/Backports) describes in detail how to work with the tooling to create backport PRs on GitHub or using the CLI tools. As described in the documentation, the [`/backport`](https://wiki.openjdk.java.net/display/SKARA/Commit+Commands#CommitCommands-/backport) command can be used to create the backport PR. If a backport PR is manually created, set the title to `Backport <original commit hash>`.
```
magicus, line 1504:
```suggestion
The Skara tooling includes support for backports. [The official Skara documentation](https://wiki.openjdk.java.net/display/SKARA/Backports) describes in detail how to work with the tooling to create backport PRs on GitHub or using the CLI tools. As described in the documentation, the [`/backport`](https://wiki.openjdk.java.net/display/SKARA/Commit+Commands#CommitCommands-/backport) command can be used on a commit (not a PR!) to create the backport PR. If a backport PR is manually created, set the title to `Backport <original commit hash>`.
```

magicus (approved):
Looks good in general. I fixed a typo for you. :) Also recommended some changes wrt Skara.

src/index.md
JesperIRL: Yes. Thank you for reviewing! Added a sentence with the label.

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
 
 After a failure is handled by excluding a test, the main JBS issue should be re-triaged and possibly given a new priority. This should be handled by the standard triage process. A test exclusion results in an outage in our testing. This outage should be taken into consideration when triaging, in addition to the impact of the bug itself.
 
+## Backing Out a Change
+
+If a change causes a regression that can't be fixed within reasonable time the best way to handle the regression can be to back out the change. Backing out means that the inverse (anti-delta) of the change is pushed to effectively undo the change in the repository. There are two parts to this task, how to do the bookkeeping in JBS, and how to do the actual backout in git or mercurial.
```
theRealAph, line 1375:
```suggestion
If a change causes a regression that can't be fixed within reasonable time, the best way to handle the regression can be to back out the change. Backing out means that the inverse (anti-delta) of the change is pushed to effectively undo the change in the repository. There are two parts to this task, how to do the bookkeeping in JBS, and how to do the actual backout in git or mercurial.
```

src/index.md
magicus: You did not like these suggestions? It has happened more than once that users try to use `/backport` on a PR, since commit commands (or even comments in general!) is not widely known about. I think at least that should be made more clear.

JesperIRL: These changes has been included in the current version. Thank you for your review and these suggestions!

magicus: I see that now. Thanks. It looked to me that they had not been included, but I must have just gotten lost in the Github UI.

RealCLanger (approved):
Looks good.

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+
+Setting the fix version of a backport to `N` is always wrong. JDK `N` has already been released (or you wouldn't be backporting to it) and can't get any more fixes.
+
+## Requesting approvals for backports
+
+In order to be allowed to push a change to one of the OpenJDK Update Release repositories, an approval is required. The [official process for how to request push approval for a backport](https://openjdk.java.net/projects/jdk-updates/approval.html) describes in detail how to work with JBS when requesting approvals. In short, there's a label `jdk<release>u-fix-request` that should be added to the main JBS issue. Also put a motivation as to why the issue needs to be backported as a comment in the main issue. Once the label and motivation has been added, wait for the maintainers of the release to approve your request. The approval will be indicated with a label, `jdk<release>u-fix-yes`, added to the main issue.
+
```
GoeLin, line 1501: Hi,
I would propose to differ the development and release
repositories most update projects use (like jdk11u-dev and jdk11u).

In this case, in the first sentence, it should say 
"... OpenJDK Update **Development** repositories ..."

and at the end you could add
  If an update is in rampdown in the Release repository, 
  urgent fixes need to be approved with
  `jdk<release>u-critical-request` / `jdk<release>u-critical-yes`.

If you want to keep it short, omitting this is ok.  But this is 
also missing on the "JDK Updates: Requesting push approval for fixes"
wiki page referred to from here.

GoeLin (changes requested):
A general remark wrt. backouts:
One thing I have always been wondering about: 
What to do if a backported issue is backed out?
For an example see JDK-8274338. It was backported
to 11.0.14 but then backed out again as JDK-8276943.
Nevertheless in JBS the original issue is marked as backported 
to 11.0.14.
Should we advise to remove this link, and change
the backport issue and the backout issue to 
bugs?
I think this would make obvious that the issue is NOT 
fixed in the older release (here 11.0.14).

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+## Using the Skara tooling to help with backports
+
+The Skara tooling includes support for backports. [The official Skara documentation](https://wiki.openjdk.java.net/display/SKARA/Backports) describes in detail how to work with the tooling to create backport PRs on GitHub or using the CLI tools. As described in the documentation, the [`/backport`](https://wiki.openjdk.java.net/display/SKARA/Commit+Commands#CommitCommands-/backport) command can be used on a commit (not a PR!) to create the backport PR. If a backport PR is manually created, set the title to `Backport <original commit hash>`.
```
jerboaa, line 1504: Suggestion: "... set the PR title to `Backport <original commit hash>`. This ensures that the the bots will recognize it as a backport as opposed to a fix specifically targetting an older release only. One can tell whether or not the bots recognized a PR as a backport by the `backport` label being added if it is being recognized."

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+
+There can be cases where it is motivated to create a backport issue before the fix is done. In these cases set the fix version of the backport to `N-pool`, where `N` is the release train the backport is targeting. E.g. `17-pool`. Please note that even if a backport issue is created ahead of time, all work done in JBS is still done in the main issue.
+
+Obviously it's possible to set the fix version to the exact release the backport is targeting, but this isn't recommended. When a change is pushed, the bots will look at the main issue as indicated in the PR title, and look for backports with the current `N.0.x` release version as fix version, if no such backport is found they will look for `N-pool`, and if that isn't found either, a new backport issue will be created. This means that if the backport has an exact fix version set, but is delayed and misses the release indicated by the fix version, a new backport issue is created with a small mess as the result.
```
gnu-andrew, line 1494: Using the exact release e.g. `openjdk8u322` is the only option with 8u.

src/index.md
gnu-andrew: +1

src/index.md
gnu-andrew: +1

gnu-andrew (changes requested):

JesperIRL:
> A general remark wrt. backouts:
> One thing I have always been wondering about:
> What to do if a backported issue is backed out?
> For an example see JDK-8274338. It was backported
> to 11.0.14 but then backed out again as JDK-8276943.
> Nevertheless in JBS the original issue is marked as backported
> to 11.0.14.
> Should we advise to remove this link, and change
> the backport issue and the backout issue to
> bugs?
> I think this would make obvious that the issue is NOT
> fixed in the older release (here 11.0.14).

This is an excellent question. I'll add some info on that topic to the section covering backouts.

JesperIRL:
> Note there is a distinction, in this context, between "backing out" a change and later reversing it, or changing it. "backouts" are when a change is pushed and quickly demonstrates it has issues and needs to be undone promptly. But sometimes a change is available for a while and we only later discover it is the source of an unexpected consequence (functional or performance) and so it needs to be modified or sometimes completely reversed. Generally speaking the former should never be backported because a backport should only happen after a change has had some bake time.

In the particular case mentioned here, JDK-8274338, the issue itself was not backed out from mainline, but only the 11.0.14 backport was backed out from 11u. This is a rare case (I hope) but clearly motivates some guidance on how to handle it as the standard model for backports is not sufficient here.

> I would think we should follow what is done for the primary issue in such a case - the backport issue should be Resolved as "Fix failed", and link to the backout issue of course.

I have done that for this backport but this is not enough. The main issue still have the backported-by-link to JDK-8276785 (11.0.14) so in the main issue it still looks like this fix was backported to 11.0.14. A minimum action here would be to also remove the backported-by-link and add a relates-to-link instead.

src/index.md
JesperIRL: Thank you for these suggestions! I have included this.

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+
+Over time it's likely that the code base will diverge between mainline and any given update release, and the cost of backporting will increase. The cost in this case is not only the effort needed to perform the actual backport, but also the cost inferred by bugs introduced by the backport. This should be taken into consideration when deciding if a change should be backported or not. For more details on how to reson around what to backport, [this email from JDK 8 Updates Project lead Andrew Haley](https://mail.openjdk.java.net/pipermail/jdk8u-dev/2020-June/012002.html) has some guidelines for JDK 8u. The reasoning in this mail is specific to JDK 8u, but will in general apply to any JDK release in maintenance mode.
```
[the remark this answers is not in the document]
JesperIRL: Nice catch! Fixed.

src/index.md
JesperIRL: Thank you for this suggestion. Included.

src/index.md
```diff
@@ -1370,6 +1370,139 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+
+::: {.box}
+[Terminology]{.boxheader}
+
+Main issue - The top issue in a backport hierarchy. Eg. [JDK-8272373](https://bugs.openjdk.java.net/browse/JDK-8272373) is a main issue, while [JDK-8277498](https://bugs.openjdk.java.net/browse/JDK-8277498) and [JDK-8277499](https://bugs.openjdk.java.net/browse/JDK-8277499) are backport issues of this main issue.
+:::
+
+In general there is no need to create backport issues in JBS manually. All work that is done in JBS in preparation for a backport (requesting approvals etc) is done in the main issue. The backport issue will be created automatically by the bots when you integrate the change to the source code repository.
+
+There can be cases where it is motivated to create a backport issue before the fix is done. In these cases set the fix version of the backport to `N-pool`, where `N` is the release train the backport is targeting. E.g. `17-pool`. Please note that even if a backport issue is created ahead of time, all work done in JBS is still done in the main issue.
```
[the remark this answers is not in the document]
JesperIRL: Thank you for the review. I have included the change and CSR example so resolving this comment. Reply about n-pool in the next comment :-)

src/index.md
JesperIRL: I don't see why there should be a difference between 8u and later releases. The tooling handle these releases exactly the same way as far as I know. Is there any official communication saying that 8u shouldn't use 8-pool?

GoeLin:
Hi,

Yes, what happened is what Jesper describes.  A change that is good in head
and seems to be fine in an older release still can have problems not noticed
during standard testing.  I think so far we saw this twice in 11u.
Besides introducing a bug in the old release, it might also turn out
that the fix changes the behavior in an undesired way, e.g. breaking the standard.

Yes, replacing the backported-by-link by relates-to is what I would propose, too.
But should we change the type of issue from "Backport" to something  else, probably "Bug"?

And yes, it must be put into this documentation in a way that it is clear that this only applies to backouts of backported issues.

Anyways, thanks for picking this up!

src/index.md
GoeLin: +1 :)

kevinrushforth:
Are you talking about changing the backport relationship of a backport record in JBS for which a fix has already been pushed, but then subsequently backed out? That seems like the wrong thing to do, given that there is in fact a commit in the update release repo with the main bug ID.

I might be missing something, but would it be sufficient to indicate fix-failed in the backport record (for the release where it was later backed out), and then file a REDO bug?

JesperIRL:
If it's desired to redo the backport filing a REDO bug might be sufficient, but if it's found to be too complex to fix whatever issues caused the backout a REDO bug won't be filed, leaving us in a situation where the main issue looks like it has been backported to 11u when in fact the change is not available in 11u.

Even if the REDO bug is filed, keeping the backported-by-link to the backed out backport would cause confusion as the main issue still looks like it's already backported. And if the second backport is pushed to the same update release there will be two backport issues with the same fix version. In this case JBS will only show one of them in the list of backports, and it's unlikely to be the correct one.

JesperIRL:
I don't think we should change the type of the backed out backport. I don't see any benefit in that, just extra work. We aren't trying to hide the fact that the backport was attempted.

kevinrushforth:
> If it's desired to redo the backport filing a REDO bug might be sufficient, but if it's found to be too complex to fix whatever issues caused the backout a REDO bug won't be filed, leaving us in a situation where the main issue looks like it has been backported to 11u when in fact the change is not available in 11u.

That's a good point: Without a redo bug it will be confusing to anyone looking at it. It's also might be a little confusing to remove the record, but perhaps not as confusing as leaving it there.

> Even if the REDO bug is filed, keeping the backported-by-link to the backed out backport would cause confusion as the main issue still looks like it's already backported. And if the second backport is pushed to the same update release there will be two backport issues with the same fix version.

The redo will use a different main bug IDs, so I don't see that this adds confusion. Your earlier point is probably enough of a reason to do what you suggest.

Btw, I wouldn't suggest deleting the backport relationship for a bug that has been backed out in mainline (shortly) after it was backported, and thus backed out in the update release as well. What you propose makes sense only for bugs that remain in mainline but not in one or more of the update releases. This should be very rare for the reasons you mentioned earlier, but it can happen.

JesperIRL:
> The redo will use a different main bug IDs, so I don't see that this adds confusion. Your earlier point is probably enough of a reason to do what you suggest.

I would expect the redo issue in this case to be a new backport issue from the same main issue.

> Btw, I wouldn't suggest deleting the backport relationship for a bug that has been backed out in mainline (shortly) after it was backported, and thus backed out in the update release as well. What you propose makes sense only for bugs that remain in mainline but not in one or more of the update releases. This should be very rare for the reasons you mentioned earlier, but it can happen.

Right, I totally agree.

kevinrushforth:
>> The redo will use a different main bug IDs, so I don't see that this adds confusion. Your earlier point is probably enough of a reason to do what you suggest.

> I would expect the redo issue in this case to be a new backport issue from the same main issue

That violates our best practices of not reusing a bug. This would lead to two commits in the repo with the same bug ID and different content. There's a reason we don't ever do this in the mainline.

JesperIRL:
> That violates our best practices of not reusing a bug. This would lead to two commits in the repo with the same bug ID and different content. There's a reason we don't ever do this in the mainline.

Right.. So what is the best solution here then?

1. Create the redo issue of type Bug and push it as a separate fix to 11u. Link the new issue with the old main issue as related. Pro: New bugid is a Bug. Cons: The old main issue won't show the new fix as a backport and won't have 11u in the list of backports.
2. Create the redo issue of type Bug and push it as a separate fix to 11u. Link the new issue as a backport of the old main issue. Pro: New bugid is a Bug, old main issue will show the new bug as a backport. Cons: Backport-link to an issue of type Bug.
3. Create the redo issue of type Backport and push it with the redo bugid. Pro: Everything will look good in JBS. Cons: Bugid in change history will refer to a backport issue.
4. Create a new mainline Bug issue and close it as duplicate of the old main issue. Create the redo issue as a backport of the duplicate and push with the duplicate bugid. Pro: Most things will probably look ok, won't break any rules I think. Cons: Seems like  a lot of process for the sake of process, not a very pragmatic solution.

I'd vote for #3 or some better solution not listed here.

GoeLin:
> I don't think we should change the type of the backed out backport issue. I don't see any benefit in that, just extra work. We aren't trying to hide the fact that the backport was attempted.

ok.

GoeLin:
> Btw, I wouldn't suggest deleting the backport relationship for a bug that has been backed out in mainline (shortly) after it was backported, and thus backed out in the update release as well. What you propose makes sense only for bugs that remain in mainline but not in one or more of the update releases. This should be very rare for the reasons you mentioned earlier, but it can happen.
Yes. This issue is not a problem, as the backout would be backported, too, and then again an eventual redo.

GoeLin:
> Right.. So what is the best solution here then?

Is having twice the same bugId such a big problem?
The bugid anyways does not refer to the JBS issue connected to the push.
And these backouts are very rare.

I would propose to remove the backport link and then just do a new backport.  This is the most simple approach for the developers working on the backports. Would this work with skara? Would the repeated bugId break any tooling?

kevinrushforth:
I prefer option 1, but Jesper is right about the con of doing that.

Option 2 seems messy, although it might work.

Option 3 won't work since Skara forbids using a backport ID in a pull request (and I would not be in favor of changing that).

Option 4 seems too heavy-weight and error prone.

> I would propose to remove the backport link and then just do a new backport. This is the most simple approach for the developers working on the backports. Would this work with skara? Would the repeated bugId break any tooling?

Let's call this option 5 then. This won't break any tooling, and from a JBS point of view, won't be confusing. While I still don't much like the idea of having two commits in a repo with the same commit message but with different contents, this seems the most expedient approach. And as noted this should be very rare.

GoeLin:
I think JBS should make transparent in which older release a bug was actually fixed by having a proper backported relation.

> However, those cases should be "the same" fix
We could redo the broken fix as-is with the same bug-id and have the backported relation. Then we could push the adaption to the target release as a follow-up bug.

GoeLin:
Hi, 
sorry, this moved out of my mind.
This seems to be a complicated issue, I don't want to block this PR with it.
Maybe it should be a follow up.
As all the other content is very useful, so we should get it commited.

GoeLin (approved):
LGTM

irisclark (approved):

PR #70, opened 2021-12-17
Add start on a section about logging 

LudwikJaniuk:
It's hard to find information on how one should write logging code in Hotspot. This PR is only a first attempt at pointing the reader in the right direction, but a lot could be added in the future:
 * In which contexts can one not use unified logging?
 * How much should one log?
 * What other logging strategies are there?

---------
### Progress

### Reviewers
 * [Jesper Wilhelmsson](https://openjdk.java.net/census#jwilhelm) (@JesperIRL - **Reviewer**) ⚠️ Review applies to 0c5c333e6d3b2678bfde40281f54659656201393

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/guide pull/70/head:pull/70` \
`$ git checkout pull/70`

`$ git checkout pull/70` \
`$ git pull https://git.openjdk.java.net/guide pull/70/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 70`

`$ git pr show -t 70`

</details>

<a href="https://git.openjdk.java.net/guide/pull/70.diff">https://git.openjdk.java.net/guide/pull/70.diff</a>

</details>

### DIFF
--- a/src/index.md
+++ b/src/index.md
@@ -1317,7 +1317,6 @@ After a failure is handled by excluding a test, the main JBS issue should be re-
 * [JEP 158: Unified JVM Logging](https://openjdk.java.net/jeps/158)
 :::
 
-
 While developing your fix, your might want your code to output some diagnostic information. You might even want to leave some logging output on in the code you check in, to facilitate future diagnostics. 
 
 The appropriate way to print logging output from Hotspot is through the [Unified Logging Framework (JEP 158)](https://openjdk.java.net/jeps/158). It gives you a lot of nice features and enables common command-line options for all logging. Messages can also be "decorated" with e.g. uptime, level, tags. The JEP contains a thorough description of the feature, but a quick example might look like:
@@ -1343,6 +1342,13 @@ log_<level>(Tag1[,...])(fmtstr, ...)
 At the time of writing, the different log levels can be found in `src/hotspot/share/logging/log.hpp`. 
 
 # Working with the legacy Mercurial servers
+
+::: {.box}
+[Quick Links]{.boxheader}
+
+* [Mercurial: The Definitive Guide](http://hgbook.red-bean.com/)
+* [OpenJDK Mercurial Server](https://hg.openjdk.java.net/)
+:::
 
 After the initial release of the JDK source code into OpenJDK in 2007 the OpenJDK project moved from TeamWare to using Mercurial. Starting in 2019 the source revision control has been moved to Git and GitHub. Even though most large projects have moved to Git by now, some still use the Mercurial servers. To access these projects some additional setup is required.
 
--- a/src/index.md
+++ b/src/index.md
@@ -1307,7 +1307,7 @@ The fix for the main issue should remove the test from the ProblemList or remove
 
 After a failure is handled by excluding a test, the main JBS issue should be re-triaged and possibly given a new priority. This should be handled by the standard triage process. A test exclusion results in an outage in our testing. This outage should be taken into consideration when triaging, in addition to the impact of the bug itself.
 
-# Adding new code in Hotspot
+# HotSpot development
 
 ## Logging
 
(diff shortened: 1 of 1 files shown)


src/index.md
```diff
@@ -1307,6 +1307,40 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+
+## Logging
+
+::: {.box}
+[Quick Links]{.boxheader}
+
+* [JEP 158: Unified JVM Logging](https://openjdk.java.net/jeps/158)
+:::
+
+While developing your fix, your might want your code to output some diagnostic information. You might even want to leave some logging output on in the code you check in, to facilitate future diagnostics.
```
JesperIRL, line 1320: "output on in" -> "output in" ?

src/index.md
```diff
@@ -1307,6 +1307,40 @@ The fix for the main issue should remove the test from the ProblemList or remove
 
 After a failure is handled by excluding a test, the main JBS issue should be re-triaged and possibly given a new priority. This should be handled by the standard triage process. A test exclusion results in an outage in our testing. This outage should be taken into consideration when triaging, in addition to the impact of the bug itself.
 
+# Adding new code in Hotspot
```
JesperIRL, line 1310: HotSpot should be written with capital S. More occurrences later.

I wonder if we should call this section something more generic, like "HotSpot Development" so that it can host more than just addition of code.

src/index.md
```diff
@@ -1307,6 +1307,40 @@ The fix for the main issue should remove the test from the ProblemList or remove
...
+-Xlog:gc*
+~~~
+
+The API should be similar to:
+
+~~~c++
+log_<level>(Tag1[,...])(fmtstr, ...)
+~~~
+
+At the time of writing, the different log levels can be found in `src/hotspot/share/logging/log.hpp`.
```
JesperIRL, line 1342: Please link the source file to the GitHub source.

JesperIRL (changes requested):
Thank you for adding this! This is a good start of an important section.

JesperIRL:
> Sorry but this hotspot specific development detail seems completely out of place in this guide. There should be a hotspot development guide where this would be a valid section - but right now we only have (some out-dated) wikis on hotspot development.

I think area specific content can be part of the Developers' Guide. There's already some, and I've gotten several inquiries to add more on the specific rules around area specific practice and process. I like this section as it's a first step towards actual development guidance. Most of the guide has been focused on process so far.

dholmes-ora:
I think the Developer's Guide should be focused on process. It is a guide to enable people to make contributions. It cannot be a technical development guide for all different areas of the JDK - that would be completely impractical and would lead to factoring out the real "getting started" process oriented guide.

There is a place for a Hotspot Developer's Guide to replace the wiki materials, but that should be a completely separate document IMO.

David

JesperIRL:
I agree that once there is enough material to motivate a separate document, we should separate development and process into two different documents. And once there's enough material to motivate separating HotSpot and libraries development we can do that as well. At the moment I don't think there's enough to motivate that new document though. Even if we move the build section as well, that's pretty much it - building and a brief HotSpot logging.

If more people volunteer to update and move information from the wiki that could clearly happen pretty soon though.

LudwikJaniuk:
The fact is that there isn't a hotspot specific developer guide as of now. What's important to me is getting stuff like this documented *some place* a new member will find it. It's outrageous to come aboard as a new JVM developer, and have access to no documented answers to such basic necessities as logging, and just be expected to figure it out from hearsay. 

So could we just get it into the documentation-base as a first step, and leave discussions about perfect locations as step 2?

LudwikJaniuk:
@dholmes-ora while the dev guide focuses largely on process, I don't think it was ever meant to exclusively cover that. In the first few paragraphs I read: 

> The OpenJDK Community is a fairly pragmatic place. “Do the right thing” is most often the right course of action. Still, if people do things in the same right way then everyone’s work becomes more transparent and easier for others to follow. For this reason most parts of the development process have standard flows that are the recommended ways to do things.

> The goal of this guide is to answer questions that developers of the JDK might have around development process, tooling, standards, and so forth.

Logging is a question developers have. And we do have a "right way to do the thing": Unified Logging! So it seems well-aligned with the goals of the dev guide to answer that question and point people in the right direction.

JesperIRL:
I agree with Ludvig here, the guide has never been strictly about process only. Guidelines for how to clone the source code and examples of shell scripts to help development are very hands-on examples of content that has been in the Guide since the first version.

I'm going to allow this section. If it turns out that we have a lot of area specific content that makes the Guide difficult to overview we can (and should) split it into separate documents later.

dholmes-ora:
If the Developer's Guide tries to incorporate hotspot specific development/programming discussions that have nothing at all to do with general processes and are relevant only to hotspot developers, then the guide is going to be too overwhelming and harder to use for the rest of the OpenJDK Developer community IMO. YMMV.

PR #94, opened 2023-01-13
Clearly state that users should create branch

magicus:
A recent PR submission to the JDK made me aware that the guide does not strongly enough emphasize the importance of creating a separate branch. 

(If you do put your changes directly in `master`, the Skara bots will put up harsh warnings, which is not a nice way to welcome someone who has tried to do everything right, including reading the guide)

---------
### Progress

### Reviewers
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - Author) ⚠️ Review applies to [aa8e02c6](https://git.openjdk.org/guide/pull/94/files/aa8e02c6ff4f7fafbf26244baf2086a723ebc524)
 * [Alexey Ivanov](https://openjdk.org/census#aivanov) (@aivanov-jdk - no project role)
 * [Jesper Wilhelmsson](https://openjdk.org/census#jwilhelm) (@JesperIRL - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/guide pull/94/head:pull/94` \
`$ git checkout pull/94`

`$ git checkout pull/94` \
`$ git pull https://git.openjdk.org/guide pull/94/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 94`

`$ git pr show -t 94`

</details>

<a href="https://git.openjdk.org/guide/pull/94.diff">https://git.openjdk.org/guide/pull/94.diff</a>

</details>


dholmes-ora (approved):
Seems quite reasonable.

Thanks.

src/guide/cloning-the-jdk.md
```diff
@@ -45,7 +45,7 @@ Pushes to your personal fork can be made either using HTTPS or SSH. These exampl
...
 
-The recommendation is to always create a new branch for any change you intend to implement. By doing that you can easily work on many different changes in parallel in the same code repository. Unless you know what you are doing, the recommendation is also to always base your new branch on the `master` branch.
+You should always create a new branch for any change you intend to implement. If your PR gets accepted, it will be squashed and pushed by the OpenJDK bots. This mean that if you make changes to your `master` branch, it will diverge from the upstream `master` branch. This will cause a lot of trouble and make any future contributions messy and complicated. Having a separate branch for each change also means that you can easily work on many different changes in parallel in the same code repository. Unless you know what you are doing, the recommendation is also to always base your new branch on the `master` branch.
```
aivanov-jdk, line 48:
```suggestion
You should always create a new branch for any change you intend to implement. If your PR gets accepted, it will be squashed and pushed by the OpenJDK bots. This means that if you make changes to your `master` branch, it will diverge from the upstream `master` branch. This will cause a lot of trouble and make any future contributions messy and complicated. Having a separate branch for each change also means that you can easily work on many different changes in parallel in the same code repository. Unless you know what you are doing, the recommendation is also to always base your new branch on the `master` branch.
```

aivanov-jdk (commented):
A small typo to be addressed.

src/guide/cloning-the-jdk.md
magicus: Thanks!

aivanov-jdk (approved):

JesperIRL (changes requested):
"should" is still just a recommendation. If you want to make this authoritative "must" is a better choice.

"messy and complicated" is very subjective. Could it be described more clearly what will actually be the problem?

magicus:
If you have ever tried this, you'd discover what "messy and complicated" means. Your HEAD will diverge from the HEAD of upstream, which will cause a wide variety of different issues depending on what you are trying to do with your fork. I could not reasonably list them all, nor will it serve a purpose since it will only distract from the point which is "don't do this".

I still think the recommendation is a "should", not a "must". It is not that things do not work if you do otherwise, it is just that you will make things needlessly complicated for yourself. Something that is a "must" is more like the JBS/PR title match, there are no room for alternatives there, no matter how savvy you are.

magicus:
But sure, I can try to rewrite this to sound less subjective, while still keeping out the gory details (which do not really matter at this point).

JesperIRL (approved):
Yes, I like this much better.

aivanov-jdk (approved):

PR #79, opened 2022-04-06
Section on PRs

JesperIRL:
Some tips for creating pull requests.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change must be properly reviewed

### Reviewers
 * [Iris Clark](https://openjdk.java.net/census#iris) (@irisclark - **Reviewer**) ⚠️ Review applies to 67993f287ddd0bff38c04cf4da26c82e30807ad7
 * [Phil Race](https://openjdk.java.net/census#prr) (@prrace - Author) ⚠️ Review applies to 67993f287ddd0bff38c04cf4da26c82e30807ad7

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/guide pull/79/head:pull/79` \
`$ git checkout pull/79`

Update a local copy of the PR: \
`$ git checkout pull/79` \
`$ git pull https://git.openjdk.java.net/guide pull/79/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 79`

View PR using the GUI difftool: \
`$ git pr show -t 79`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/guide/pull/79.diff">https://git.openjdk.java.net/guide/pull/79.diff</a>

</details>

PR #98, opened 2023-02-20
More on testing

JesperIRL:
* Added clarifications and expectations around testing and review
* Moved GHA section (unchanged)
* Removed outdated stuff from the glossary

---------
### Progress

### Reviewers
 * [Iris Clark](https://openjdk.org/census#iris) (@irisclark - **Reviewer**) ⚠️ Review applies to [1a938bf5](https://git.openjdk.org/guide/pull/98/files/1a938bf5802163dcda4213f966d0fe7637264e42)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/guide.git pull/98/head:pull/98` \
`$ git checkout pull/98`

`$ git checkout pull/98` \
`$ git pull https://git.openjdk.org/guide.git pull/98/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 98`

`$ git pr show -t 98`

</details>

<a href="https://git.openjdk.org/guide/pull/98.diff">https://git.openjdk.org/guide/pull/98.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/guide/pull/98#issuecomment-1437209684)


src/guide/hotspot-development.md
```diff
@@ -1,5 +1,10 @@
 # HotSpot Development
 
+See [Working With Pull Requests](#working-with-pull-requests) for generic guidance and requirements around pushing changes. For the HotSpot codebase there are a few additional requirements:
+
+* Your change must have been approved by two Committers out of which at least one is also a Reviewer
```
kimbarrett, line 5: The original requirement, from the Oracle-internal wiki ("Hotspot Code Review Guidelines"), was "At least two reviewers (one of which is a Reviewer)".  The "two Committers" requirement suggested here (taken from the openjdk wiki) was a quiet change introduced when the openjdk wiki page was created based on the internal wiki.  In earlier (private) discussion this change was questioned.  We want to encourage thoughtful and knowledgeable reviews even from folks who are new or only lightly connected to the project and so don't have Committer status (yet), rather than effectively saying such reviews don't count for anything in the official process.

src/guide/hotspot-development.md
```diff
@@ -1,5 +1,10 @@
...
+* Your change must have passed through HotSpot tier 1 testing with zero failures (See tier1 definition in `test/hotspot/jtreg/TEST.groups`.)
+
```
kimbarrett, line 7: I thought the 24 hour rule was part of the HotSpot tailoring (present in both the internal and external HotSpot review guidelines), rather than a JDK-wide requirement.  But it's in the generic requirements in the Guide.  Am I misremembering, or is that a change I didn't notice before?

kimbarrett: And I don't see the 24 hour rule being strictly adhered to elsewhere.  For example, I see the occasional quick turnaround in core-libs changes.

src/guide/hotspot-development.md
dholmes-ora: I agree - please keep this as "two reviewers".

src/guide/hotspot-development.md
dholmes-ora: I think the 24 hour rule was hotspot specific. Other areas do very quick turn-arounds.

src/guide/testing-the-jdk.md
```diff
@@ -167,6 +169,16 @@ make test TEST=gtest:$X/$variant
 
 The second example above runs tests which match the regexp `$X.*` on a specific variant of the JVM. The variant is one of client, server, etc.
 
+## GitHub actions
+
+GitHub has a feature called **GitHub actions** (GHA) that can be used to automate testing. The GHA is executed whenever a push is made to a branch in your repository. The bots will show the results of the GHA in your PR when you create or update it. The GHA in the JDK project is configured to run a set of tests that is commonly known as **tier 1**. This is a relatively fast, small set of tests that tries to verify that your change didn't break the JDK completely. In tier 1 the JDK is built on a small set of platforms including (but not necessarily limited to) Linux, Windows, and MacOS, and a few tests are executed using these builds.
```
dholmes-ora, line 174: GHA also does a whole bunch of cross-compilation builds for PPC, ARM and S390.

src/guide/hotspot-development.md
JesperIRL: I can't say that I have studied what happens in reality in other areas, but the 24 hour text was added as a general rule in "Working with pull requests" in PR #79 which is one of the more reviewed PRs in the Guide. Several developers from all areas of the JDK was involved and no one had any issues with the rule. In fact, the only comment about that particular guidance was from a person active in the libraries area and he suggested to add the sentence about holidays etc to say that sometimes even 24 hours isn't enough. So it seems to me that this is an accepted rule in general. Wether people actually follow the rule may be a different story.

src/guide/hotspot-development.md
JesperIRL: Agreed and fixed.

src/guide/testing-the-jdk.md
JesperIRL: Added a sentence about this.

JesperIRL:
@kimbarrett @dholmes-ora Thank you for your reviews!

src/guide/hotspot-development.md
seanjmullan: FWIW, I think the 24 hour "rule" is a good one for all component teams, except I think it should be more like a best practice, rather than a hard rule. Exceptions would be for trivial fixes and fixes for build errors or tier 1 or tier 2 failures where a quicker turnaround is necessary.

irisclark: I think the current version of the new text is fine given that the requirements are preceded by the phrase "For the HotSpot codebase...".

irisclark (approved):

src/guide/working-with-pull-requests.md
```diff
@@ -24,7 +24,7 @@ Creating the PR is essentially the same as asking a large group of people to sta
 
 * Is the copyright statement at the top of each modified source file correct?
 
-* Did you run all relevant tests on the final version of the change? (Yes, I mean final! If you only knew how many times people _only changed a comment_ and caused a build failure.)
+* Did you run all relevant tests on the final version of the change? **Yes, I mean final!** If you knew how many times people _only changed a comment_ and caused a build failure you wouldn't question this.
```
stuart-marks, line 27: I agree with the sentiment here, but the tone is a little odd. It's one of only a couple places in the Guide where it refers to itself in the first person. When I run across wording like this I immediately think, "Who is the author?" In writing such as a blog post or newspaper column, there is a by-line, and so it's clear who is speaking. I don't think that's the case here, so it's just kind of strange.

The point is worth making though, so I'd suggest substituting something like,

> Did you run a full build and all relevant tests on the **final** version of the change? (It's important to do this on the truly final version, as even an apparently trivial change in a comment can break the build.)

The other use of first person is in testing-the-jdk.md; I'll make a comment there.

src/guide/testing-the-jdk.md
```diff
@@ -167,6 +169,16 @@ make test TEST=gtest:$X/$variant
...
+
+Please keep in mind that the tier 1 tests run by the GHA should only be seen as a smoke test that finds the most critical breakages, like build errors or if the JDK is DOA. These tests can never replace the targeted testing that you always must do on your changes. There are several areas of the JDK that aren't part of tier 1 at all. To see exactly what tier 1 includes, please see the various TEST.groups files that you will find in the subdirectories of [`jdk/test/`](https://github.com/openjdk/jdk/tree/master/test).
+
 ## Excluding a test
 
 Sometimes tests break. It could be e.g. due to bugs in the test itself, due to changed functionality in the code that the test is testing, or changes in the environment where the test is executed. While working on a fix, it can be useful to stop the test from being executed in everyone else's testing to reduce noise, especially if the test is expected to fail for more than a day. There are two ways to stop a test from being run in standard test runs: ProblemListing and using the `@ignore` keyword. Removing tests isn't the standard way to remove a failure. A failing test is often a regression and should ideally be handled with high urgency.
```
stuart-marks, line 184: The next paragraph begins,

> I'll say it right away so that it's not forgotten at the end:

This is an odd first person wording here. In this case the introductory clause doesn't add anything so I'd suggest simply removing it.

JesperIRL: Fixed.

PR #101, opened 2023-04-21
Things to consider before PR

JesperIRL:
Text about what kind of changes that may or may not be desired in OpenJDK. Based on @prrace's text on https://openjdk.org/groups/client-libs/

---------
### Progress

### Reviewers
 * [Phil Race](https://openjdk.org/census#prr) (@prrace - Author) ⚠️ Review applies to [683fc260](https://git.openjdk.org/guide/pull/101/files/683fc26082263d0d8b825d33020706f7484c92eb)
 * [Iris Clark](https://openjdk.org/census#iris) (@irisclark - **Reviewer**) ⚠️ Review applies to [2c2a0390](https://git.openjdk.org/guide/pull/101/files/2c2a03909d3f144fda0119d03b4762eb231d1a73)
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - no project role) ⚠️ Review applies to [2c2a0390](https://git.openjdk.org/guide/pull/101/files/2c2a03909d3f144fda0119d03b4762eb231d1a73)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/guide.git pull/101/head:pull/101` \
`$ git checkout pull/101`

`$ git checkout pull/101` \
`$ git pull https://git.openjdk.org/guide.git pull/101/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 101`

`$ git pr show -t 101`

</details>

<a href="https://git.openjdk.org/guide/pull/101.diff">https://git.openjdk.org/guide/pull/101.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/guide/pull/101#issuecomment-1518062434)


RogerRiggs:
I like the tone of this advice, to be careful about what is proposed an thorough in the preparation.
Possibly there can be guidance about being open to comments on the change and accepting of the recommendations including not pursuing it.

prrace:
> I like the tone of this advice, to be careful about what is proposed and thorough in the preparation.
> Possibly there can be guidance about being open to comments on the change and accepting of the recommendations including not pursuing it.

I agree on both counts, if by the 2nd comment you imply that if you decide the advice doesn't apply to you, others might not agree :-)

Regardless, this text is a good addition to the guide.

prrace (approved):
No objection to an additional comment as suggested by Roger, but regardless looks good.

irisclark (approved):
Looks good.  I echo Roger and Phil's comment regarding accepting recommendations.  Also, sunk cost isn't a reason to make a change.

desruisseaux:
In the _"What is the main intention of your change?"_ section, there is no item for _"Reducing code size"_. Would it be an item worth to add? For example in `java.awt.image`, there is:

* Blocks of code repeated in different places, which could have been factored in package-private methods.
* Unnecessary codes overwriting what the default values already provided, for example:
  *  loops for setting all elements to 0 in a newly created array,
  * setting non-final fields to zero in the constructor,
  * invoking constructor with arguments when there is a no-argument constructor providing the desired values.

Should the guide mentions a policy for potential "reducing code size" tasks?

stuart-marks:
> **What is the main intention of your change?**

I think we need more emphasis on this point. What I usually do is ask, "What problem are you trying to solve?" The important thing here is to understand the problem itself, independent of any solution (and independently of the solution that currently happens to be in the PR). A number of other questions and lines of thought emanate from thinking about the problem. For example: is this the _right_ problem to solve? Does solving this problem create worse problems elsewhere; that is, is there a net benefit? Does this problem need to be solved in the JDK, or can and should it be solved elsewhere (e.g., in tooling)?

stuart-marks:
An item that would be good to add to the list of things to consider is maintainability. On more than one occasion I've seen people propose changes that added a lot of complexity for not much benefit. Or, the proposal was a "quick hack" that solved an immediate problem but that increased complexity or technical debt. Maybe this is the right tradeoff, or maybe it's not; it needs to be carefully considered. A helpful idea to keep in mind is that changes accepted into the JDK will need to be maintained by other people in perpetuity.

stuart-marks:
Tradeoffs around performance-related proposals. We all want the JDK to be faster right? There are a bunch of tradeoffs around proposals that attempt to improve the performance of the system. One is that a change might speed up an infrequently occurring case but slow down a frequently occurring case. That's usually the wrong tradeoff.

Another tradeoff is between code complexity and performance. Code needs to be maintained in perpetuity (see item on maintainability) so adding to code complexity for a modest or negligible performance improvement is often a bad trade.

Finally, it's when considering performance proposals, they often need to be backed up by benchmarks. It's important that one understand benchmark results properly. In particular, one needs to pay attention to the error ranges and memory allocation results.

JesperIRL:
> In the "What is the main intention of your change?" section, there is no item for "Reducing code size".

@desruisseaux Isn't this covered in the Refactoring/Cleanup bullet?

desruisseaux:
> > In the "What is the main intention of your change?" section, there is no item for "Reducing code size".

> Isn't this covered in the Refactoring/Cleanup bullet?

The Refactoring/Cleanup bullet seems to be about aesthetic improvements, which is difficult to quantify. Reducing the code size as a measurable benefit. Whether such benefit is worth a change, this is the question.

JesperIRL:
> Possibly there can be guidance about being open to comments on the change and accepting of the recommendations including not pursuing it.

@RogerRiggs I've added some sentences about this in the PR-section where the review is covered.

JesperIRL:
> What I usually do is ask, "What problem are you trying to solve?"

Added this, used your words. Thanks!

> An item that would be good to add to the list of things to consider is maintainability.

This is covered in Why is my change rejected?. I can't really fit it into the list of thing to consider since I don't see anyone answering "maintainability" to the intention question. If they do it's probably the same as Refactoring/Cleanup.

> Tradeoffs around performance-related proposals.

Added a few sentences on this.

JesperIRL:
> The Refactoring/Cleanup bullet seems to be about aesthetic improvements, which is difficult to quantify. Reducing the code size as a measurable benefit. Whether such benefit is worth a change, this is the question.

Ok. I've added a few words to clarify that reducing code size is included in Refactoring/Cleanup. Will that cover your examples?

desruisseaux:
> Ok. I've added a few words to clarify that reducing code size is included in Refactoring/Cleanup. Will that cover your examples?

It is okay for me, it is not very important. I do not know if I'm the only one to think that "reducing footprint" and "improve maintainability" would have been an answer to the _"what will a cleanup actually achieve?"_ question.

On the maintainability aspect, the reason is that when relatively large method bodies are duplicated (e.g. [here](https://github.com/openjdk/jdk/blob/406021ad585eec1ec007535ed6b08c2ebffec2ee/src/java.desktop/share/classes/java/awt/image/Raster.java#L1037) versus [here](https://github.com/openjdk/jdk/blob/406021ad585eec1ec007535ed6b08c2ebffec2ee/src/java.desktop/share/classes/java/awt/image/Raster.java#L1145) where only the last line differs), if a change is applied in one method, there is a risk that the duplicated method is forgotten.

src/guide/working-with-pull-requests.md
```diff
@@ -14,6 +14,8 @@ All code reviews in OpenJDK are done in public. Once you open your PR for public
 
 It's also worth taking the extra time to see if the change can be split into a few different separate changes. A large change will take more effort and thus attract fewer Reviewers. Smaller changes will get reviewed faster and get better quality reviews. You can compare proposing a single large change to proposing ten individual small unrelated changes. What happens in practice when all these ten changes are presented as one PR is that there's a focus on say 5-6 of these smaller changes and no one really looks hard at the other 4-5. For complexity, even small changes that are hard to understand and test may be risky.
 
+The timing of your change will also affect the availability of reviewers. The JDK runs on a [six-months release cadence](#release-cycle). During the months around the start of the ramp down phase most area experts will be bussy working on their own changes and reviewing major features that are planned for the release. If you propose a change during this period (basically May-June, or November-December) it may take a long time before you get the required reviews.
```
RogerRiggs, line 17: typo "bussy" -> "busy'

RogerRiggs (approved):

src/guide/contributing-to-an-open-jdk-project.md
```diff
@@ -8,6 +8,30 @@ If you have a success story where Java solved your problem, or if you successful
 
 If you're in a position to choose what programming language to use in a project, in a tutorial, or in a class, you have the power to enlarge the Java community in a very direct way, and your colleagues or students will get an opportunity to learn one of the most used programming languages in the world.
 
+## Things to consider before proposing changes to OpenJDK code
+
+Every change in JDK code carries risk of changes in behavior which adversely affect applications. Generally we're looking to improve the functionality and capability and sometimes performance of the platform without that negative consequence. We do have many thousands of tests, but they're inevitably incomplete in coverage. So we need to ask ourselves whether each change is worthwhile - and some may not be no matter how well intentioned.
```
seanjmullan, line 13: This sounds like all changes in behavior are negative, which is not true. I suggest inserting "may" before "adversely".

seanjmullan, line 13: > We do have many thousands of tests, but they're inevitably incomplete in coverage.

This sentence needs more explanation, I think. If you are changing code, ideally you would provide tests that test that behavior change if the current tests are inadequate. Does this sentence add anything to the main points in this paragraph - otherwise, maybe it should be removed.

irisclark (approved):

src/guide/working-with-pull-requests.md
JesperIRL: Fixed.

src/guide/contributing-to-an-open-jdk-project.md
JesperIRL: Fixed.

JesperIRL: Agreed. Removed.

JesperIRL:
Thank you all for your reviews!
/integrate

PR #118, opened 2024-01-12
Clarifications and X-wnf label

JesperIRL:
Fixed a bug, cleaned up a few JBS field references, and added the X-wnf label to the dictionary.

---------
### Progress

### Reviewers
 * [Iris Clark](https://openjdk.org/census#iris) (@irisclark - **Reviewer**) ⚠️ Review applies to [37ed8eed](https://git.openjdk.org/guide/pull/118/files/37ed8eed47b8f9d4f4869b193ac875f58d6dc69f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/guide.git pull/118/head:pull/118` \
`$ git checkout pull/118`

`$ git checkout pull/118` \
`$ git pull https://git.openjdk.org/guide.git pull/118/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 118`

`$ git pr show -t 118`

</details>

<a href="https://git.openjdk.org/guide/pull/118.diff">https://git.openjdk.org/guide/pull/118.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/guide/pull/118#issuecomment-1889646241)


irisclark (approved):

src/guide/jbs-jdk-bug-system.md
```diff
@@ -46,7 +46,7 @@ A few things to keep in mind when filing an issue:
   * If you find a similar issue that was closed as [Cannot Reproduce]{.jbs-value} then it may be appropriate to re-open that one - if you don't have direct access to JBS you can file a bug using the [Bug Report Tool](https://bugreport.java.com/) requesting that it be reopened.
 * Make a reasonable attempt to narrow down which build or release the failure first appeared in.
 * Set [Affects Version/s]{.jbs-field} to the earliest JDK version(s) where the failure was seen.
-  * If the failure is found in an update train of the JDK (e.g. 11.0.x), if possible please see if it's also present in [mainline](https://github.com/openjdk/jdk).
+  * If the failure is found in an update train of the JDK (e.g. 11.0.x), please see (if possible) if it's also present in [mainline](https://github.com/openjdk/jdk).
```
dholmes-ora, line 49: Not sure the proposed change makes any grammatical difference. But the bigger question is how they should reflect any findings on mainline in the "affects version" field? This is case where I would definitely want to see the latest version known to fail as well as the earliest.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -345,7 +345,17 @@ This table contains some frequently used JBS labels and their meaning. Please he
   <tr>
     <td class="dictionary">[*(Rel)*[-na]{.jbs-label}]{#rel-na}</td>
     <td class="dictionary">
-      The [Affects Version/s]{.jbs-field} field is used to indicate the releases where an issue has been seen - it's implied that the issue is also applicable to newer releases. Where this isn't the case - for instance a bug in code that was removed in *(Rel)* - then use the *(Rel)-na* to indicate this. Note that there should only be **one** *(Rel)*[-na]{.jbs-label} label on any JBS issue.
+      Labels of the form [11-na]{.jbs-label} or [21-na]{.jbs-label} should be used when a bug is not applicable to a **more recent** release family.
+
+      The [Affects Version/s]{.jbs-field} field is used to indicate the releases where an issue has been seen - it's implied that the issue is also applicable to newer releases. Where this isn't the case - for instance a bug in code that was removed in *(Rel)* - then use *(Rel)*[-na]{.jbs-label} to indicate this. Note that there should only be **one** *(Rel)*[-na]{.jbs-label} label on any JBS issue.
```
dholmes-ora, line 350: Take the case documented above where a bug is found in 11.x and we ask they check if it affects mainline. If it doesn't then they add, e.g. `23-na`, but this tells us nothing about when it was actually fixed or disappeared. If someone checks 21 and find the bug is not there then shouldn't `21-na` also be added? Or should it replace `23-na` - with the assumption that `<rel>-na` implies `<rel+N>-na` for all `N`?

src/guide/jbs-jdk-bug-system.md
```diff
@@ -345,7 +345,17 @@ This table contains some frequently used JBS labels and their meaning. Please he
...
+
+      Don't use the label to indicate that a bug isn't relevant to an earlier release. If a bug is not present in an earlier release you can use [Introduced in Version]{.jbs-field} to indicate where the bug was introduced. If a bug is present in an earlier release, but shouldn't be fixed there, use [*(Rel)*[-wnf]{.jbs-label}]{#rel-wnf}.
```
dholmes-ora, line 352: But from a backporting perspective I, for example, would want to know if the issue is present in 8u, and I can easily test that. But I don't know if the issue in present in 9 or 10 and I would not want to waste time testing those versions just so I can set the "Introduced In" field. It makes more sense to me, to use `8-na` to indicate that JDK 8 has been checked and has no issue. Alternatively, the guide should be directing me to perhaps create an explicit backport record for 8u and close it as "Not an Issue".

JesperIRL: The intention is to have an upper boundary on the need to fix an issue - the X-na label, and a lower boundary - the X-wnf label. A label that is used to spot-indicate specific releases without saying anything about the before or after is much less useful when trying to automate things. For a human looking at a single bug what you say makes sense, but for a tool it would be impossible to determine if 8-na means just 8 or 8 and later.

The labels are slightly badly named which is why this discussion exists. But this is what we currently have. To change the names of the labels is a much larger discussion.

src/guide/jbs-jdk-bug-system.md
JesperIRL: Added a sentence to indicate that adding the mainline version may be a good idea.

src/guide/jbs-jdk-bug-system.md
JesperIRL: The label should be updated (replaced) in that case. I have tried to clarify this in the text.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -347,9 +347,11 @@ This table contains some frequently used JBS labels and their meaning. Please he
...
+      The [Affects Version/s]{.jbs-field} field is used to indicate the releases where an issue has been seen - it's implied that the issue is also applicable to newer releases. Where this isn't the case - for instance a bug in code that was removed in *(Rel)* - then use *(Rel)*[-na]{.jbs-label} to indicate this. Note that there should only be **one** *(Rel)*[-na]{.jbs-label} label on any JBS issue. The label indicates the first known release where the issue is not present. If it's found that the issue is not present in an earlier release, change the label to indicate this.
 
-      Don't use the label to indicate that a bug isn't relevant to an earlier release. If a bug is not present in an earlier release you can use [Introduced in Version]{.jbs-field} to indicate where the bug was introduced. If a bug is present in an earlier release, but shouldn't be fixed there, use [*(Rel)*[-wnf]{.jbs-label}]{#rel-wnf}.
+      It's not recommended to specify update releases like 17u4 or 21u in the label. Please stick to labels like [17-na]{.jbs-label} and [21-na]{.jbs-label}.
```
dholmes-ora, line 352: Again this can get a bit grey. Suppose an issue is present in 17 but not in 21+, and the change in 21 is then backported to 17uN. In that case putting 17-na would be very misleading as the bug is present in 17 (and 17uM M<N)

JesperIRL: Not recommended doesn't mean it's forbidden. There will always be edge cases where what makes the most sense isn't necessarily the recommended way.

PR #139, opened 2024-12-03
Addition of a causes link type

calnan:
There is a proposal for the addition of a causes/caused by link type.  Tied into that there is now a separate section which covers the main link types that we use

---------
### Progress

### Reviewers
 * [Jesper Wilhelmsson](https://openjdk.org/census#jwilhelm) (@JesperIRL - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/guide.git pull/139/head:pull/139` \
`$ git checkout pull/139`

`$ git checkout pull/139` \
`$ git pull https://git.openjdk.org/guide.git pull/139/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 139`

`$ git pr show -t 139`

</details>

<a href="https://git.openjdk.org/guide/pull/139.diff">https://git.openjdk.org/guide/pull/139.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/guide/pull/139#issuecomment-2515738885)
</details>


src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
 * The [Description]{.jbs-field} usually explains what went wrong and how the failure was found, then there's some investigation and eventually the root cause is found. At this point the [Summary]{.jbs-field} should be updated to correctly describe the bug. The [Description]{.jbs-field} however should remain a description of how the failure was found.
 * The [Affects Version/s]{.jbs-field} should be updated if you in your investigation finds that the issue is older than what is indicated by the current [Affects Version/s]{.jbs-field}.
 
+### Linking Issues
+
+An important aspect of any issue is making clear how it is connected/related to other issues, this linking is done when closing an issue - such as marking an issue as a duplicate and [adding a duplicate link](#closing-issues-as-duplicates) or when using the ‘Create Backport’ or ‘Create CSR’ options are used which we automatically create ‘backported by’ or ‘csr for’ links (never create an issue of type Backport or CSR by hand). The other use of linking is to associate an issue with one or more other issues that share some relation with the issue and is done during the triage or resolution of the issue.
```
dholmes-ora, line 257: Linking can occur at any stage of an issue's lifecycle: when it is filed, when it is triaged, when it is being worked upon, or when it is being closed. Suggestion:
> An important aspect of any issue is making clear how it is connected/related to other issues. This can occur at any stage of the issue's lifecycle. For example, as information becomes available that might suggest a cause, or similar issue; or when a Backport or CSR request is created; or when closing as a duplicate of another issue.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+There are the following link types:
+
+’duplicate of` - See [Closing issues as duplicates] for more information.
```
dholmes-ora, line 261:
```suggestion
’duplicate of` - Normally set automatically - see [Closing issues as duplicates] for more information.
```

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+‘backported by’ - When creating a backport with the “More -> Create Backport” option a link is automatically created between the main issue and the new Backport
```
dholmes-ora, line 263:
```suggestion
‘backported by’ - Normally set automatically when creating a backport with the “More -> Create Backport” option
```

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+‘CSR for’ - When creating a CSR with the “More -> Create CSR” option a link is automatically created between the main issue and the new CSR
```
dholmes-ora, line 265:
```suggestion
‘CSR for’ - Normally created automatically when creating a CSR with the “More -> Create CSR” option a link
```

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+‘blocks’ - when a fix is broken down into a number of parts the ‘blocks’ link should be used to ensure they are all fixed before the main issue is considered resolved - see [implementing large changes](#implementing-large-changes]
```
dholmes-ora, line 267:
```suggestion
‘blocks’ - For when other issues are dependent on the current issue being resolved/fixed before they can be. For example, when a fix is broken down into a number of parts the ‘blocks’ link should be used to ensure they are all fixed before the main issue is considered resolved - see [implementing large changes](#implementing-large-changes]
```

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+‘relates to’ - there are no rules as to when or why to create a relates link apart from not duplicating an existing “duplicated by”, ‘backported by’, ‘csr for’ or ‘blocked by’ links. In general, you should link any other issue that has a bearing on the situation where you feel the related issue should be reviewed in order to have a better understanding of what is going on
```
dholmes-ora, line 269: I'd also caution against excessive linking with "relates to" - for example, if a test fails you should not link every other issue with which this test has been previously been associated just because the same test was involved. The "relates to" issue should have some significance in relation to the cause and/or fix, for the current issue.

Note that it is often the case that a "relates to" situation is later determined to be a "duplicates" situation, in which case you have to manually delete the "relates to link" after Closing. (I wonder if JBS could automate that?)

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+‘causes’/‘caused by’ - the causes link implies a stronger relationship than relates. If an issue B is filed which can be traced back to the fix for issue A then ‘A causes B’ (or ‘B was caused by A’).  A ‘causes’ link would be added to A if after the integration or release of A it is found that additional work needs to be done. This might be that extra work in another area forgotten and needs to be completed or the more common case would be that a fix ‘causes’ a change of behavior (intentional or otherwise).  For a regression, if you identify the fix that caused it, add a caused-by link to that issue and set the Introduced in Build or Introduced in Version field.  Whenever looking to backport a fix the developer should look for both ‘blocked by’ and ‘causes’ links in order to understand the set of fixes that should be backported. Likewise, if A has already been backported the new causes linked issues will need to be assessed to see if it is important enough to be backported as well. On B make sure to set the [Introduced in Build]{.jbs-field} or [Introduced in Version]{.jbs-field} field.
```
dholmes-ora, line 271: It seems somewhat disproportionate to explain `caused by` in so much depth compared to the others. The backporting notes really belong elsewhere, not in this definition list. I would suggest a simpler and more succinct definition:
> 'causes'/'caused-by' - For when the current issue can be directly attributed to a code change that was integrated under another issue. When such a cause is known you can also set the 'Introduced in Version' and 'Introduced in Build' fields.

dholmes-ora: I also think it will be common for a "relates" link to turn into a "cause" link, in which case we need general instructions about removing multiple links to the same issue. Ideally JBS would let us edit a link and change its type.

dholmes-ora (commented):
NIce to see such a section being added - and I've missed a "caused-by" link for many years (why did we not have it from day one?). A few suggestions however.

Thanks

src/guide/jbs-jdk-bug-system.md
calnan: done
I think we may want to emphasis that backports should not be created by hand - but perhaps that should be in the backport section - I'll look at that along with your last point

src/guide/jbs-jdk-bug-system.md
calnan: done

src/guide/jbs-jdk-bug-system.md
calnan: done

src/guide/jbs-jdk-bug-system.md
calnan: done

src/guide/jbs-jdk-bug-system.md
calnan: done

dholmes-ora (commented):
Thanks for updates.

src/guide/jbs-jdk-bug-system.md
JornVernee: In some cases a fix can reveal an existing bug, e.g. because it makes an unlikely issue more likely, or because it added more tests which then find an existing bug. From the sound of it, just because a failure started to occur after commit `xyz`, doesn't necessarily mean that that commit 'causes' the issue? Some explicit guidance on whether to apply this link type in such cases would be useful I think.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -228,12 +228,12 @@ Now that the issue is in the right component and has the basic information, the
...
    * Changes that don't affect product code, but are only against documentation: [[noreg-doc]{.jbs-label}](#noreg-doc)
    * Well contained issues that seem to be easy to fix: [[starter]{.jbs-label}](#starter)
    * Enhancements that are pure cleanups: [[cleanup]{.jbs-label}](#cleanup)
    * Project specific issues usually have their own labels as well
+1. Managing regressions - for a bug (B) where behavior has _incorrectly_ changed from a previous fix (A) sure that the label [[regression]{.jbs-label}](#regression) is added.  Once it is known what fix caused the regression a 'caused by' link should be added to 'B' or a causes link to 'A'. A ‘causes’ link would be added to A if after the integration or release of A it is found that additional work needs to be done. This might be that extra work in another area forgotten and needs to be completed or the more common case would be that a fix ‘causes’ a change of behavior (intentional or otherwise).  If 'A' has been identifed as well as a caused-by link to that issue and set the [Introduced in Build]{.jbs-field} and [Introduced in Version]{.jbs-field} fields of 'B', based on which release 'A' was fixed in.
```
aivanov-jdk, line 236:
```suggestion
1. Managing regressions - for a bug (B) where behavior has _incorrectly_ changed from a previous fix (A) ensure that the label [[regression]{.jbs-label}](#regression) is added.  Once it is known what fix caused the regression a ’caused by’ link should be added to 'B' or a ‘causes’ link to 'A'. A ‘causes’ link would be added to A if after the integration or release of A it is found that additional work needs to be done. This might be that extra work in another area forgotten and needs to be completed or the more common case would be that a fix ‘causes’ a change of behavior (intentional or otherwise).  If 'A' has been identified as well as a caused-by link to that issue and set the [Introduced in Build]{.jbs-field} and [Introduced in Version]{.jbs-field} fields of 'B', based on which release 'A' was fixed in.  
```

In “…a ’caused by’ link should be added to 'B' or a ‘causes’ link to 'A'”, the word “causes” needs the single-quotes to refer to the type of the link. (Also for consistency with other instances.)

In “…that a fix _‘causes’_ a change of behavior…”, the word “causes” should not have the quotes because the verb is used in its literal meaning rather being a type.

I find it really hard to understand the last two sentences in this paragraph. “This might be that extra work in another area was forgotten…” — is “was” missing here? The part after “or”: if I understand correctly, _“the more common case”_ is a clarification, wrapping it the commas, dashes or parentheses could be easier to understand. However, it it's the most common case, should it be listed first, instead?

In the last sentence, “If 'A' has been identified as well as a caused-by link…” what does “as well as” compare to? Is it just “as”? And the link type should have the single quotes around it and not have the hyphen “…as a _‘caused by’_ link…”.

Using italics for marking up the link type (in addition to the single quotes) could clarify where the text refers to a type rather than uses the verbs literally.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
 
-* For a regression, if you identify the fix that caused it, add a relates-to link to that issue (and add a [[regression_]{.jbs-label}(ID)](#regression_id) label) and set the [Introduced in Build]{.jbs-field} or [Introduced in Version]{.jbs-field} field.
 * The [Description]{.jbs-field} usually explains what went wrong and how the failure was found, then there's some investigation and eventually the root cause is found. At this point the [Summary]{.jbs-field} should be updated to correctly describe the bug. The [Description]{.jbs-field} however should remain a description of how the failure was found.
 * The [Affects Version/s]{.jbs-field} should be updated if you in your investigation finds that the issue is older than what is indicated by the current [Affects Version/s]{.jbs-field}.
 
+### Linking Issues
+
+An important aspect of any issue is making clear how it is connected/related to other issues. This can occur at any stage of the issue's lifecycle. For example, as information becomes available that might suggest a cause, or similar issue (relates to); or when a Backport or CSR request is created; or when closing as a duplicate of another issue.
```
aivanov-jdk, line 257:
```suggestion
An important aspect of any issue is making clear how it is connected/related to other issues. This can occur at any stage of the issue's lifecycle. For example, as information becomes available that might suggest a cause, or similar issue (relates to).
```
This should be enough as an example. The following list presents what link types are available and their meaning.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+There are the following link types:
+
+’duplicate of` - Normally set automatically - see [Closing issues as duplicates] for more information
```
aivanov-jdk, line 261:
```suggestion
‘duplicate of’ - Normally set automatically - see [Closing issues as duplicates] for more information
```

Does markdown support a definition list? Should we use HTML `<dl>` and `<dt>`, `<dd>` tags to mark up the list?

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+‘backported by’ - Normally set automatically when creating a backport with the “More -> Create Backport” option
```
aivanov-jdk, line 263: Or when the backport is integrated?

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
 * The [Description]{.jbs-field} usually explains what went wrong and how the failure was found, then there's some investigation and eventually the root cause is found. At this point the [Summary]{.jbs-field} should be updated to correctly describe the bug. The [Description]{.jbs-field} however should remain a description of how the failure was found.
 * The [Affects Version/s]{.jbs-field} should be updated if you in your investigation finds that the issue is older than what is indicated by the current [Affects Version/s]{.jbs-field}.
 
+### Linking Issues
+
+An important aspect of any issue is making clear how it is connected/related to other issues. This can occur at any stage of the issue's lifecycle. For example, as information becomes available that might suggest a cause, or similar issue (relates to); or when a Backport or CSR request is created; or when closing as a duplicate of another issue.
+
+There are the following link types:
```
aivanov-jdk, line 259: I suggest marking up the issue type with bold `**` if not with `<dt>` and using a descriptive sentence maybe, for example:

> **‘duplicate of’** is normally set automatically when another issue is closed as a _‘duplicate’_, see \[Closing issues as duplicates\] for more information.
> 
> **‘backported by’** is set automatically when a fix is pushed to an older release or when a backport is created manually using the **More** -> **Create Backport** option.

The list would be easier to read if the description of all the link types followed the same pattern.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+‘backported by’ - Normally set automatically when creating a backport with the “More -> Create Backport” option
+
+‘CSR for’ - When creating a CSR with the “More -> Create CSR” option a link is automatically created between the main issue and the new CSR
+
+‘blocks’ - For when other issues are dependent on the current issue being resolved/fixed before they can be. For example, when a fix is broken down into a number of parts the ‘blocks’ link should be used to ensure they are all fixed before the main issue is considered resolved - see [implementing large changes](#implementing-large-changes]
+
+‘relates to’ - there are no rules as to when or why to create a relates link apart from not duplicating an existing “duplicated by”, ‘backported by’, ‘csr for’ or ‘blocked by’ links. In general, you should link any other issue that has a bearing on the situation where you feel the related issue should be reviewed in order to have a better understanding of what is going on
+
+‘causes’/‘caused by’ - the causes link implies a stronger relationship than relates. If an issue B is filed which can be traced back to the fix for issue A then ‘A causes B’ (or ‘B was caused by A’)
```
aivanov-jdk, line 271:
```suggestion
‘causes’/‘caused by’ - the ‘causes’ link implies a stronger relationship than ‘relates to’. If an issue 'B' is traced back to the fix for issue 'A' then ‘A causes B’ (or ‘B was caused by A’)
```
Should it be ‘B _is_ caused by A’?

In the above text about regressions, issue A and B have dumb single quotes around them. This should be consistent here, therefore I added them in the suggestion.

aivanov-jdk (changes requested):

magicus:
How should the causes/caused by link be used when a fix is backed out, and then redone? Since this is a bit of a tricky albeit somewhat common situation, maybe we should clarify it in the guide?

src/guide/jbs-jdk-bug-system.md
calnan: reworded the section to make it clearer that relates links should have 'some significance' - agreed it would be useful if JIRA could update the relates link if it later used as a duplicate.  Will see if there is a way to spot this so that they might be manually cleaned up

src/guide/jbs-jdk-bug-system.md
calnan: refactored the text so that it now in more appropriate sections
Jorn - added a sentence "Triaging an Issue" section to cover your point - please review

src/guide/backporting.md
```diff
@@ -29,6 +29,8 @@ If, for instance, there are other changes between the original one and the follo
 
 Testing each individual change is more likely to find issues than just testing the single merged change. It's also easier and less error prone to use the `/backport` command on each commit instead of manually cherrypick and deal with the merges etc.
 
+Whenever looking to backport a fix the developer should look for both ‘blocked by’ and ‘causes’ links in order to understand the set of fixes that should be backported. Likewise, if A has already been backported the new causes linked issues will need to be assessed to see if it is important enough to be backported as well.
```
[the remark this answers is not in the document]
calnan: checked with Jesper and have changed them all to [name]{.jbs-value}

src/guide/jbs-jdk-bug-system.md
calnan: restructured a bit

src/guide/jbs-jdk-bug-system.md
calnan: added a mention of Skara

src/guide/jbs-jdk-bug-system.md
calnan: should now be consistent

src/guide/jbs-jdk-bug-system.md
calnan: I think the past-tense makes sense, have added the quotes

calnan:
I don't think the fact that fix 'A' is backed out changes the 'A causes B' link itself, although B would be closed once the back-out happens but the link would remain.
Lets see if David/Jesper have some suggestions

JesperIRL:
> I don't think the fact that fix 'A' is backed out changes the 'A causes B' link itself, although B would be closed once the back-out happens but the link would remain. Lets see if David/Jesper have some suggestions

I agree. This would be alternative 2 in step 3 of https://openjdk.org/guide/#backing-out-a-change - backing out the change is fixing the regression. We don't remove the links when issues are fixed.

magicus:
So if I have an old checkin A that caused a regression in e.g. a prior release that was not noticed at the time, and make a fix for it in B, but it is broken and is backed out in C, and then redone in D.

A causes B, B causes C, and A causes D?

dholmes-ora:
> B causes C

I don't think it makes sense to have a backout issue be marked as "caused by" the original issue. I don't think that is the sense of "cause" this was intended to capture.

magicus:
Why not? I assume that one thing you want to capture is when backporting. So if B is broken, you definitely would want to know that it caused undesired behavior that was fixed by C. Or am I mistaken?

aivanov-jdk:
> Why not? I assume that one thing you want to capture is when backporting. So if B is broken, you definitely would want to know that it caused undesired behavior that was fixed by C. Or am I mistaken?

I agree. The reason for backout is that the fix was broken, so C was caused by B.

When backporting, B and C can be skipped, therefore a ‘causes’ link from A to D makes sense to me.

magicus:
> When backporting, B and C can be skipped, therefore a ‘causes’ link from A to D makes sense to me.

But that only makes sense once D is created. When creating B, the author genuinely believes that it fixes A. It is just that later on, B was found to be broken. Should we at that point go and remove the `causes` link for A?

I'm not saying I have the right answer here, but I definitely think that whatever we come up with should be documented in the guide, since this is indeed a tricky scenario.

aivanov-jdk:
> > When backporting, B and C can be skipped, therefore a ‘causes’ link from A to D makes sense to me.
> 
> But that only makes sense once D is created. When creating B, the author genuinely believes that it fixes A. It is just that later on, B was found to be broken. Should we at that point go and remove the `causes` link for A?

I don't think we should. The relationship between A and B remains even after C is created, and after it's integrated, too.

Additional comments could be added to clarify the relationships. However, the fact that B causes C and that C is a backout highlights that B isn't actually fixed. For this reason, D should be created as soon as it's clear B will be backed out — at the same time when C is created.

> I'm not saying I have the right answer here, but I definitely think that whatever we come up with should be documented in the guide, since this is indeed a tricky scenario.

I don't know the right answer either.

Yet I think A should have the _‘causes’_ link to both B and D, possibly a _‘relates to’_ link to C too.

magicus:
... and B causes C, right?

I don't think we need to explain every edge case, but I think the specific case of a backout could be served by being explicit in what to do in that situation.

dholmes-ora:
> Why not? I assume that one thing you want to capture is when backporting. So if B is broken, you definitely would want to know that it caused undesired behavior that was fixed by C. Or am I mistaken?

First off if B is broken it should be Verified as Fix Failed and that is what tells backporters there is an issue. It then has a link to C (the backout) and a link to a REDO issue (potentially). These links are evident regardless of how the link is described.

If we say that the Backout issue is caused by B then that would imply the REDO issue is also caused by B - they are both only necessary because B was broken. Are you proposing that as well? If so the Backout section of the guide also needs to reflect this.

I understand that without B being broken there is no need for a Backout issue but that relationship is to me more specific than simply causal. Maybe we need a backed-out-by kind of link.

src/guide/jbs-jdk-bug-system.md
calnan: not at the moment it seems in commonmark - https://talk.commonmark.org/t/description-list/289/41

src/guide/jbs-jdk-bug-system.md
calnan: updated

calnan:
I'm not seeing that the last set of comments warrant any changes to the text
Jesper handing over to you to review

JesperIRL (approved):

src/guide/backporting.md
```diff
@@ -29,6 +29,8 @@ If, for instance, there are other changes between the original one and the follo
 
 Testing each individual change is more likely to find issues than just testing the single merged change. It's also easier and less error prone to use the `/backport` command on each commit instead of manually cherrypick and deal with the merges etc.
 
+Whenever looking to backport a fix the developer should look for both [blocked by]{.jbs-value} and [causes]{.jbs-value} links in order to understand the set of fixes that should be backported. Likewise, if A has already been backported the new [causes]{.jbs-value} linked issues will need to be assessed to see if it is important enough to be backported as well.
```
aivanov-jdk, line 32:
```suggestion
Whenever looking to backport a fix, the developer should look for both [blocked by]{.jbs-value} and [causes]{.jbs-value} links in order to understand the set of fixes that should be backported. Likewise, if A has already been backported, the new [causes]{.jbs-value} linked issues will need to be assessed to see if it is important enough to be backported as well.
```

src/guide/jbs-jdk-bug-system.md
```diff
@@ -228,12 +228,12 @@ Now that the issue is in the right component and has the basic information, the
...
    * Well contained issues that seem to be easy to fix: [[starter]{.jbs-label}](#starter)
    * Enhancements that are pure cleanups: [[cleanup]{.jbs-label}](#cleanup)
    * Project specific issues usually have their own labels as well
+1. Managing regressions - for a bug (B) where behavior has _incorrectly_ changed from a previous fix (A) ensure that the label [[regression]{.jbs-label}](#regression) is added.  Once it is known what fix caused the regression a [caused by]{.jbs-value} link should be added to 'B' or a [causes]{.jbs-value} link to 'A'. A [causes]{.jbs-value} link would also be added to A if the fix _causes_ a change of behavior (intentional or otherwise) or it is found after integration, that additional work needs to be done. If 'A' has been identified as well as a adding a [caused by]{.jbs-value} link to that issue, set the [Introduced in Build]{.jbs-field} and [Introduced in Version]{.jbs-field} fields of 'B', based on which release 'A' was fixed in.  Do not add a [caused by]{.jbs-value} link if there was no specific product fix which _caused it_, for example, the addition of a test which finds an underlying problem should not be linked.
```
aivanov-jdk, line 236:
```suggestion
1. Managing regressions - for a bug (B) where behavior has _incorrectly_ changed from a previous fix (A), ensure that the label [[regression]{.jbs-label}](#regression) is added.  Once it is known what fix caused the regression a [caused by]{.jbs-value} link should be added to 'B' or a [causes]{.jbs-value} link to 'A'. A [causes]{.jbs-value} link would also be added to 'A' if the fix _causes_ a change of behavior (intentional or otherwise) or it is found after integration, that additional work needs to be done. If 'A' has been identified as well as a adding a [caused by]{.jbs-value} link to that issue, set the [Introduced in Build]{.jbs-field} and [Introduced in Version]{.jbs-field} fields of 'B', based on which release 'A' was fixed in.  Do not add a [caused by]{.jbs-value} link if there was no specific product fix which _caused it_, for example, the addition of a test which finds an underlying problem should not be linked.
```

“If 'A' has been identified as well as” — identified as what? It looks as if something's missing…  
Would it be better to say, “~~as well as~~ in addition to adding a [caused by]{.jbs-value} link to that issue, set the…?

This way is clearer, in my opinion.

I'd move the _Introduced in Version_ field before _Introduced in Build_ as the version is more important.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+An important aspect of any issue is making clear how it is connected/related to other issues. This can occur at any stage of the issue's lifecycle. For example, as information becomes available that might suggest a cause, or similar issue (relates to).
+
+There are the following link types:
+
+[duplicate of]{.jbs-value} - Normally set automatically - see [Closing issues as duplicates] for more information
+
+[backported by]{.jbs-value} - Normally set automatically when creating a backport with the “More -> Create Backport” option, or by the Skara tooling
+
+[CSR for]{.jbs-value} - When creating a CSR with the “More -> Create CSR” option a link is automatically created between the main issue and the new CSR
```
aivanov-jdk, line 265:
```suggestion
[CSR for]{.jbs-value} - When creating a CSR with the “More -> Create CSR” option, a link is automatically created between the main issue and the new CSR
```

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+[blocks]{.jbs-value} - For when other issues are dependent on the current issue being resolved/fixed before they can be. For example, when a fix is broken down into a number of parts the [blocks]{.jbs-value} link should be used to ensure they are all fixed before the main issue is considered resolved - see [implementing large changes](#implementing-large-changes]
+
+[relates to]{.jbs-value} - To avoid lots of relates links, the links should have some significance in relation to the cause and/or fix, for the current issue. In addition, relates links should not duplicate an existing [duplicated by]{.jbs-value}, [backported by]{.jbs-value}, [csr for]{.jbs-value} or [blocked by]{.jbs-value} link.  In particular, it may be necessary to manually remove a [relates to]{.jbs-value} link if it is later added as a [duplicated by]{.jbs-value} or [caused by]{.jbs-value} link
```
aivanov-jdk, line 269:
```suggestion
[relates to]{.jbs-value} - To avoid lots of relates links, the links should have some significance in relation to the cause and/or fix for the current issue. In addition, relates links should not duplicate an existing [duplicated by]{.jbs-value}, [backported by]{.jbs-value}, [csr for]{.jbs-value} or [blocked by]{.jbs-value} link.  In particular, it may be necessary to manually remove a [relates to]{.jbs-value} link if it is later added as a [duplicated by]{.jbs-value} or [caused by]{.jbs-value} link
```

The comma in the end of the first sentence is not need, is it?

Should “relates links” be replaced with “relates to links” in the first two sentences? An italics could be used to mark them up to make it clear that these words refer to a type of the link.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -249,10 +249,27 @@ Once you are made, or you make yourself, the assignee of an issue you take on th
...
+
+[causes]{.jbs-value}/[caused by]{.jbs-value} - the causes link implies a stronger relationship than relates. If an issue 'B' can be traced back to the fix for issue 'A' then ‘A causes B’ (or ‘B was caused by A’)
```
aivanov-jdk, line 271:
```suggestion
[causes]{.jbs-value}/[caused by]{.jbs-value} - the _causes_ link implies a stronger relationship than _relates_. If an issue 'B' can be traced back to the fix for issue 'A' then ‘A causes B’ (or ‘B is caused by A’)
```

Mark up the type of the link; use the same tense for both *‘A causes B’* and *‘B is caused by A’*.

aivanov-jdk (changes requested):

src/guide/jbs-jdk-bug-system.md
calnan: the comma is fine, updated the first relates

src/guide/jbs-jdk-bug-system.md
calnan: made the changes

src/guide/jbs-jdk-bug-system.md
calnan: done

calnan:
Jesper, as discussed I moved the order of the paragraphs in "Backporting multiple related changes"

src/guide/jbs-jdk-bug-system.md
calnan: done

JesperIRL (approved):

PR #145, opened 2025-03-03
Re-open when info received

JesperIRL:
If issues are left as Resolved no-one will notice the new information.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace

### Reviewers
 * [Iris Clark](https://openjdk.org/census#iris) (@irisclark - **Reviewer**) ⚠️ Review applies to [07a5e1af](https://git.openjdk.org/guide/pull/145/files/07a5e1af4a23cb26bc2226e5c4b0d6dd3d9de17f)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/guide.git pull/145/head:pull/145` \
`$ git checkout pull/145`

Update a local copy of the PR: \
`$ git checkout pull/145` \
`$ git pull https://git.openjdk.org/guide.git pull/145/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 145`

View PR using the GUI difftool: \
`$ git pr show -t 145`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/guide/pull/145.diff">https://git.openjdk.org/guide/pull/145.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/guide/pull/145#issuecomment-2695655043)
</details>

PR #146, opened 2025-03-14
No localization in RDP2

JesperIRL:
Localization can't be expected for changes made in RDP2.

@irisclark This change also contain your fixes from #145 that I apparently didn't commit before integrating.


irisclark (approved):

src/guide/jbs-jdk-bug-system.md
```diff
@@ -337,7 +337,7 @@ If it's determined that an issue has been fixed, but it's unknown what change th
 
 ### Closing incomplete issues
 
-As mentioned above, issues that lack the information needed to investigate the problem are placed in status [Resolved]{.jbs-value} - [Incomplete]{.jbs-value}. Triage teams should monitor incomplete issues in their area and if needed ping the relevant person. When new information has been received, the bug should be put back to status [Open]{.jbs-value}. If the required information hasn't been obtained within reasonable time (3-4 weeks) the bug should be closed as [Incomplete]{.jbs-value}.
+As mentioned above, issues that lack the information needed to investigate the problem are placed in status [Resolved]{.jbs-value} - [Incomplete]{.jbs-value}. Triage teams should monitor incomplete issues in their area and if needed ping the relevant person. When new information is received, the bug should be returned to status [Open]{.jbs-value}. If the required information hasn't been obtained within reasonable time (3-4 weeks) the bug should be closed as [Incomplete]{.jbs-value}.
```
aivanov-jdk, line 340: Does “…the bug should be closed…” imply that the *Resolved* status changes to *Closed*?

src/guide/jbs-jdk-bug-system.md
```diff
@@ -431,10 +431,10 @@ This table contains some frequently used JBS labels and their meaning. Please he
                            [*(Rel)*[-fix-yes]{.jbs-label}]{#rel-fix-yes}<br />
                            [*(Rel)*[-fix-no]{.jbs-label}]{#rel-fix-no}</td>
     <td class="dictionary">
-      Used in ramp down phase 2 to indicate that an issue would be of interest to get integrated into release *(Rel)*. E.g., [jdk12u-fix-request]{.jbs-label}
+      *(Rel)*[-fix-request]{.jbs-label} is used in ramp down phase 2 to indicate that an issue would be of interest to be integrated into release *(Rel)*. E.g., [jdk12-fix-request]{.jbs-label}
```
aivanov-jdk, line 434: Should the section be expanded for update releases? The labels like `jdk12u-fix-request` and `jdk24u-fix-request` are constantly used to [request approval for backporting](https://openjdk.org/guide/#requesting-approvals-for-backports) an update release.

The `jdk##-fix-request` and `jdk##u-fix-request` are similar, but they have a different role in the OpenJDK Project.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -431,10 +431,10 @@ This table contains some frequently used JBS labels and their meaning. Please he
...
 
-      *(Rel)*[-fix-SQE-ok]{.jbs-label} is used to indicate that the issue will be covered by the test plan for *(Rel)*. E.g., [jdk12u-fix-SQE-ok]{.jbs-label}<br />
-      *(Rel)*[-fix-yes]{.jbs-label} and *(Rel)*[-fix-no]{.jbs-label} are used to indicate wether an issue has been approved for backport to *(Rel)*. E.g., [jdk12u-fix-yes]{.jbs-label}
+      *(Rel)*[-fix-SQE-ok]{.jbs-label} is used to indicate that the issue will be covered by the test plan for *(Rel)*. E.g., [jdk12-fix-SQE-ok]{.jbs-label}<br />
+      *(Rel)*[-fix-yes]{.jbs-label} and *(Rel)*[-fix-no]{.jbs-label} are used to indicate wether an issue has been approved for backport to *(Rel)*. E.g., [jdk12-fix-yes]{.jbs-label}
```
aivanov-jdk, line 437: Yet `jdk12-fix-yes` isn't applicable to backports, it's for post RDP2 integrations (which are also *backports* since the initial fix goes to mainline which is at *Rel+1*). Maybe a different wording should be used to clarify…

src/guide/making-a-change.md
```diff
@@ -68,6 +68,12 @@ If you create a new file, copy the license text from a nearby file. Do not add c
 
 If you move code from an existing file to a new file, bring the entire copyright + license header over to the new file.
 
+## Changes Late in the Release
+
+You should always aim to get your changes integrated as early in a release as possible to maximize the amount of testing done before release. If you risk running late it's almost always a better idea to defer the change to the next release rather than trying to squeeze it in in the last minute. The next release is only six months out, the world can wait even though you are eager.
```
aivanov-jdk, line 73:
```suggestion
You should always aim to get your changes integrated as early in a release as possible to maximize the amount of testing done before release. If you risk running late, it's almost always a better idea to defer the change to the next release rather than trying to squeeze it in in the last minute. The next release is only six months away, the world can wait even though you are eager.
```

A comma after the conditional cause which comes before the main clause helps reading.

I think “away” is correct when you're referring to an upcoming release; whereas “out” would imply an already released version six months ago.

*“…to squeeze it in in the last minute”* could be hard to read because of two adjacent ‘in’s.

src/guide/making-a-change.md
```diff
@@ -68,6 +68,12 @@ If you create a new file, copy the license text from a nearby file. Do not add c
...
+
+Integrating during the rampdown phase is not recommended. The bar to integrate a change in the rampdown phase is considerably higher than during the normal development phase. See [The JDK Release Process] for more information. Also note that changes that are made late in the release shouldn't expect localization since this in itself takes some time.
```
aivanov-jdk, line 75:
```suggestion
Integrating during the rampdown phase is not recommended. The bar to integrate a change in the rampdown phase is considerably higher than during the normal development phase. See [The JDK Release Process] for more information. Also note changes that are made late in the release shouldn't expect localization since this in itself takes some time.
```

Is the last sentence easier to read and understand if you omit the relative conjunction ‘that’?

src/guide/jbs-jdk-bug-system.md
JesperIRL: Yes.

src/guide/jbs-jdk-bug-system.md
JesperIRL: I'd be happy to include such an update. Could you write the text, either here in a comment or as a separate change?

src/guide/jbs-jdk-bug-system.md
JesperIRL: As you say, I think this part would benefit from clarifying what goes for the updates and for mainline.

src/guide/making-a-change.md
JesperIRL: I don't think so. Your suggestion would require a comma after "note", and I don't think that sentence would be easier to read. It could be simplified by removing "that are" (the second that) instead.

aivanov-jdk: > I don't think so. Your suggestion would require a comma after "note", and I don't think that sentence would be easier to read.

You're right. The sentence is easier to understand with both “that” on their own places.

> It could be simplified by removing "that are" (the second that) instead.

No, the second “that” can't be removed because it's the subject of the dependent sentence “that are made late in the release” which describes _changes_.

src/guide/jbs-jdk-bug-system.md
aivanov-jdk: Yeah, it makes sense to expand the section in another PR.

aivanov-jdk (approved):

PR #149, opened 2025-04-23
Updated AffectsVersion image

JesperIRL:
Updated image to make case 5 look more like the previous cases.
Also fixed a borken link.

### DIFF
--- a/src/guide/jbs-jdk-bug-system.md
+++ b/src/guide/jbs-jdk-bug-system.md
@@ -108,7 +108,7 @@ Don't:
 
   [Affects Version/s]{.jbs-field}: [11.0.20]{.jbs-value}, [17]{.jbs-value}
 
-  the label [8-na]{.jbs-label} would not be needed - as it doesn't have a JDK 8 release, or earlier, value in the [Affects Version/s]{.jbs-field}, it is not relevant to JDK 8. Also see [Usage of the (Rel)[-wnf]{.jbs-label} Label](#usage-of-rel-wnf-label)
+  the label [8-na]{.jbs-label} would not be needed - as it doesn't have a JDK 8 release, or earlier, value in the [Affects Version/s]{.jbs-field}, it is not relevant to JDK 8. Also see [Usage of the (Rel)[-wnf]{.jbs-label} Label](#usage-of-the-rel-wnf-label)
 - add multiple [-na]{.jbs-label} labels: you only need one, for example don't add both [9-na]{.jbs-label} and [11-na]{.jbs-label} — [9-na]{.jbs-label} implies all following releases therefore [11-na]{.jbs-label}, or [17-na]{.jbs-label} etc. are not needed.
 - It's not recommended to specify update releases like 17u4 or 21u in the label. Labels like [17-na]{.jbs-label} and [21-na]{.jbs-label} are in general enough.
 
--- a/src/images/affects_versions.svg
+++ b/src/images/affects_versions.svg
@@ -178,9 +178,9 @@
      inkscape:window-width="1932"
      inkscape:window-height="1186"
      inkscape:window-x="0"
-     inkscape:window-y="1225"
+     inkscape:window-y="25"
      inkscape:window-maximized="0"
-     inkscape:current-layer="svg1108" />
+     inkscape:current-layer="timeline" />
   <style
      id="style985"><![CDATA[
         rect {
@@ -470,7 +470,7 @@
        d="M 45, 56 L 495, 56"
        marker-end="url(#arrow)"
        id="path1070"
-       style="marker-end:url(#marker2763)" />
+       style="marker-end:url(#ExperimentalArrow)" />
     <text
        x="29"
        y="84"
@@ -482,7 +482,7 @@
        d="M 45, 86 L 495, 86"
        marker-end="url(#arrow)"
        id="path1076"
-       style="marker-end:url(#marker2763)" />
+       style="marker-end:url(#Arrow2Lend)" />
     <text
        x="29"
        y="114"
--- a/src/images/affects_versions.svg
+++ b/src/images/affects_versions.svg
@@ -1,8 +1,188 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<svg width="550" height="200"
-     version="1.1"
-     xmlns="http://www.w3.org/2000/svg">
-    <style><![CDATA[
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   width="550"
+   height="200"
+   version="1.1"
+   id="svg1108"
+   sodipodi:docname="affects_versions.svg"
+   inkscape:version="1.1.1 (c3084ef, 2021-09-22)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <defs
+     id="defs1112">
+    <marker
+       id="marker2763"
+       refX="5.0"
+       refY="3.0"
+       orient="auto-start-reverse"
+       inkscape:stockid="ExperimentalArrow"
+       inkscape:isstock="true">
+      <path
+         style="fill:context-stroke"
+         d="m 10,3 -10,3 0,-6 z"
+         id="path2761" />
+    </marker>
+    <marker
+       style="overflow:visible"
+       id="TriangleOutL"
+       refX="0.0"
+       refY="0.0"
+       orient="auto"
+       inkscape:stockid="TriangleOutL"
+       inkscape:isstock="true">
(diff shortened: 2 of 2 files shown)


irisclark (approved):

JesperIRL:
@aivanov-jdk, what program did you use to edit the SVG? I'm using Incscape but it seems to be suboptimal.

aivanov-jdk:
> @aivanov-jdk, what program did you use to edit the SVG? I'm using Incscape but it seems to be suboptimal.

I hand-coded it in IDEA, I find it relaxing and easier than to use a visual editor.

In fairness, I used Inkscape when I started to look at SVG. It's a handy tool for creating complex shapes. A few times, I drafted the image in Inkscape and then simplified the image by hand-coding.

aivanov-jdk:
@JesperIRL Does this look good to you?
https://github.com/openjdk/guide/compare/master...aivanov-jdk:dev-guide:update-affects-versions

Updated Affects Versions

I left the circles at this time. Removing the circles and making the case 5 line longer is really simple.

JesperIRL:
Ah, now I see why we were confused here :-)  (5) is talking about backports while the other cases are about whether something is relevant later or not. I extended the line to show that it originates from mainline. Maybe it would be even more clear if there is a left-arrow at the right end of (5)?

I also did an attempt at cleaning up the file. I'm not sure it's worth it for a single image on such a large page, but I'm also not against removing unnecessary characters :-)

JesperIRL:
I liked the left-arrow :-) At least now it's obvious that there is something different with (5).

irisclark (approved):
Still an improvement.

aivanov-jdk:
> (5) is talking about backports while the other cases are about whether something is relevant later or not. I extended the line to show that it originates from mainline.

This doesn't make sense…

In this example, mainline was 21 at the time the issue was fixed. Therefore, the line of Affects Versions had to end at 21 where the issue is fixed.

Then, it was decided to backport the fix to 17u (completed now); yet it was decided not to backport the fix to 11u.

> Maybe it would be even more clear if there is a left-arrow at the right end of (5)?

Sorry, but I don't think it makes things clearer… rather more confusing.

aivanov-jdk:
> I also did an attempt at cleaning up the file. I'm not sure it's worth it for a single image on such a large page, but I'm also not against removing unnecessary characters :-)

I don't think it's reasonable to clean up after editing in Inkscape.

At the same time, I'm providing the version which is still small.

https://github.com/openjdk/guide/compare/master...aivanov-jdk:dev-guide:update-affects-versions

Updated Affects Versions

src/images/affects_versions.svg
aivanov-jdk, line 1: Why is 12-na displayed below the affects-versions line? I think the text label should stay aligned to 8 on the left.

Aligning both 12-na and 11-wnf would make the image more appealing (which I did in my version). Yet it may look confusing.

Either way, bullet 5 below the image needs updating to clarify what it illustrates.

https://github.com/openjdk/guide/blob/71fbb94b0b7b3ca75869bc158805971a3fe3d3df/src/guide/jbs-jdk-bug-system.md?plain=1#L129

aivanov-jdk (changes requested):

src/images/affects_versions.svg
aivanov-jdk: @JesperIRL Yeah! I'll create a new PR.

> I can't really think of a better way to phrase bullet 5 without it becoming quite long. Do you have a suggestion?

Should we describe each example with words to make it clearer?

Let bullet 5 become longer if it makes things clear.

Sorry for my delayed reply.

irisclark (approved):

src/images/affects_versions.svg
JesperIRL: Ok. Did some more changes. The image should be fairly clean and makes more sense now with the updated descriptions, I think. My main problem with the old descriptions was that they repeated how to use the labels rather than explaining the situation in the example.

aivanov-jdk: > My main problem with the old descriptions was that they repeated how to use the labels rather than explaining the situation in the example.

Yes, that's the thing! I've been thinking about it too. Does it makes sense to expand the descriptions of the cases like this:

1. Affects Version contains 8 only which means this issue is applicable to 8 and all future releases (until it's fixed).
2. Affects Version contains 8, ~~8u40, 9 and 12~~, yet 8 already implies that these releases are affected.
3. (no change)
4. Affects Version contains 8, the issue is fixed in 11; then the `12-na` label indicates the issue is not applicable to 12 and subsequent versions.
5. Affects Version contains 8, the issue is fixed in 21 and is backported to 17; the label `11-wnf` indicates the fix will not be backported to 11 or 8.

JesperIRL: Yes, your descriptions are better. I have updated to use them with just minor edits.
I approved #153. If you integrate it before this is integrated I'll make sure to not touch the image in this PR :-)

src/guide/jbs-jdk-bug-system.md
```diff
@@ -122,11 +122,11 @@ Add a comment when adding a (Rel)[-wnf]{.jbs-label} label so that it's clear for
 
 ![Guidelines for setting [Affects Version/s]{.jbs-field}](affects_versions.svg)
 
-1. Issue relevant to JDK 8 and all future releases.
-2. No need to add additional releases as they are implied.
+1. Affects Version contains 8 only, which means this issue is applicable to 8 and all future releases (until it's fixed).
```
aivanov-jdk, line 125:
```suggestion
1. [Affects Version/s]{.jbs-field} contains 8 only, which means this issue is applicable to 8 and all future releases (until it's fixed).
```
Shall we mark-up the JBS field?

src/guide/jbs-jdk-bug-system.md
```diff
@@ -122,11 +122,11 @@ Add a comment when adding a (Rel)[-wnf]{.jbs-label} label so that it's clear for
...
+2. Affects Version contains 8, 8u40, 9, and 12, yet 8 already implies that the later releases are affected.
```
aivanov-jdk, line 126: I used strike-through to indicate 8u40, 9, and 12 aren't need, which could also be confusing. Therefore, it makes to explicitly state that 8u40, 9, and 12 should not be added to the Affects Version field.

src/guide/jbs-jdk-bug-system.md
```diff
@@ -122,11 +122,11 @@ Add a comment when adding a (Rel)[-wnf]{.jbs-label} label so that it's clear for
...
 3. Adding the occasional LTS release value is ok.
```
aivanov-jdk, line 127: We should expand the description of case (3) in a similar way:

```suggestion
3. [Affects Version/s]{.jbs-field} contains 11 and 21. Adding the occasional LTS release version (21 in this example) is fine; 17 is implicitly affected as well as versions after 21.
```

src/guide/jbs-jdk-bug-system.md
```diff
@@ -122,11 +122,11 @@ Add a comment when adding a (Rel)[-wnf]{.jbs-label} label so that it's clear for
...
-4. Use N-na to indicate that the issue is no longer relevant from that release, which could be due to the feature or platform being removed or the code being rewritten.
-5. Use N-wnf to indicate that a fix will not be backported to that release, or earlier.
+4. Affects Version contains 8 and the issue is fixed in 11. The 12-na label indicates that the issue is not applicable to 12 and subsequent versions.
```
aivanov-jdk, line 128:
```suggestion
4. Affects Version contains 8 and the issue is fixed in 11. The [12-na]{.jbs-label} label indicates that the issue is not applicable to 12 and subsequent versions.
```

src/guide/jbs-jdk-bug-system.md
```diff
@@ -122,11 +122,11 @@ Add a comment when adding a (Rel)[-wnf]{.jbs-label} label so that it's clear for
...
+5. Affects Version contains 8, the issue is fixed in 21 and is backported to 17. The label 11-wnf indicates that the fix will not be backported to 11 or 8.
```
aivanov-jdk, line 129:
```suggestion
5. Affects Version contains 8, the issue is fixed in 21 and is backported to 17. The [11-wnf]{.jbs-label} label indicates that the fix will not be backported to 11 or 8.
```
Mark up the label; I also moved the words for consistency with case 4.

src/images/affects_versions.svg
aivanov-jdk: Thanks!

I issued the `/integrate` command in that PR.

aivanov-jdk (changes requested):

src/guide/jbs-jdk-bug-system.md
JesperIRL: Right. I meant to do that but pushed in a hurry when I left the office... 😅

JesperIRL:
Ok, I made another pass over the descriptions and removed my changes to the image.

src/images/affects_versions.svg
aivanov-jdk: > Yes, your descriptions are better. I have updated to use them with just minor edits.

@JesperIRL May I get a contributor attribution?

